PR #4, opened 2022-03-30
Add basic API using Actix and Axum

Alexandcoats:
I went ahead and copied over the bare minimum type support and got a basic API running with the new libs.


.gitignore
```diff
@@ -1 +1,3 @@
 /target
+Cargo.lock
+.vscode/**/*
```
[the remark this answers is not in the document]
bingyanglin: To be consistent w/ bee.

src/api/responses.rs
```diff
@@ -0,0 +1,254 @@
...
+pub(crate) enum ListenerResponse {
+    /// Response of GET /info
+    Info {
+        name: String,
+        version: String,
+        #[serde(rename = "isHealthy")]
+        is_healthy: bool,
+    },
+    /// Response of GET /api/<keyspace>/messages/<message_id>
+    /// and GET /api/<keyspace>/transactions/<transaction_id>/included-message
```
bingyanglin, line 26: Do we use keyspace or api version?
```suggestion
    /// Response of GET /api/<api_version>/messages/<message_id>
    /// and GET /api/<api_version>/transactions/<transaction_id>/included-message
```

src/api/responses.rs
```diff
@@ -0,0 +1,254 @@
...
+        #[serde(rename = "networkId")]
+        network_id: Option<u64>,
+        #[serde(rename = "protocolVersion")]
+        protocol_version: u8,
+        #[serde(rename = "parentMessageIds")]
+        parents: Vec<String>,
+        payload: Option<Value>,
+        nonce: u64,
+    },
+    /// Response of GET /api/<keyspace>/messages/<message_id>/metadata
```
bingyanglin, line 37:
```suggestion
    /// Response of GET /api/<api_version>/messages/<message_id>/metadata
```

src/api/responses.rs
```diff
@@ -0,0 +1,254 @@
...
+        #[serde(rename = "ledgerInclusionState", skip_serializing_if = "Option::is_none")]
+        ledger_inclusion_state: Option<LedgerInclusionState>,
+        #[serde(rename = "conflictReason", skip_serializing_if = "Option::is_none")]
+        conflict_reason: Option<u8>,
+        #[serde(rename = "shouldPromote", skip_serializing_if = "Option::is_none")]
+        should_promote: Option<bool>,
+        #[serde(rename = "shouldReattach", skip_serializing_if = "Option::is_none")]
+        should_reattach: Option<bool>,
+    },
+    /// Response of GET /api/<keyspace>/messages/<message_id>/children
```
bingyanglin, line 58:
```suggestion
    /// Response of GET /api/<api_version>/messages/<message_id>/children
```

src/api/responses.rs
```diff
@@ -0,0 +1,254 @@
...
+    MessageChildren {
+        #[serde(rename = "messageId")]
+        message_id: String,
+        #[serde(rename = "maxResults")]
+        max_results: usize,
+        count: usize,
+        #[serde(rename = "childrenMessageIds")]
+        children_message_ids: Vec<String>,
+    },
+    /// Response of GET /api/<keyspace>/messages/<message_id>/children[?expanded=true]
```
bingyanglin, line 68:
```suggestion
    /// Response of GET /api/<api_version>/messages/<message_id>/children[?expanded=true]
```

src/api/responses.rs
```diff
@@ -0,0 +1,254 @@
...
+    MessageChildrenExpanded {
+        #[serde(rename = "messageId")]
+        message_id: String,
+        #[serde(rename = "maxResults")]
+        max_results: usize,
+        count: usize,
+        #[serde(rename = "childrenMessageIds")]
+        children_message_ids: Vec<Record>,
+    },
+    /// Response of GET /api/<keyspace>/messages?<index>
```
bingyanglin, line 78:
```suggestion
    /// Response of GET /api/<api_version>/messages?<index>
```

src/api/responses.rs
```diff
@@ -0,0 +1,254 @@
...
+    /// Response of GET /api/<keyspace>/messages?<index>[&expanded=true]
+    MessagesForIndexExpanded {
+        index: String,
+        #[serde(rename = "maxResults")]
+        max_results: usize,
+        count: usize,
+        #[serde(rename = "messageIds")]
+        message_ids: Vec<Record>,
+    },
+    /// Response of GET /api/<keyspace>/messages?<index>
```
bingyanglin, line 96:
```suggestion
    /// Response of GET /api/<api_version>/messages?<index>
```

src/api/responses.rs
```diff
@@ -0,0 +1,254 @@
...
+    MessagesForTag {
+        tag: String,
+        #[serde(rename = "maxResults")]
+        max_results: usize,
+        count: usize,
+        #[serde(rename = "messageIds")]
+        message_ids: Vec<String>,
+    },
+    /// Response of GET /api/<keyspace>/messages?<index>[&expanded=true]
```
bingyanglin, line 105:
```suggestion
    /// Response of GET /api/<api_version>/messages?<index>[&expanded=true]
```

src/api/responses.rs
```diff
@@ -0,0 +1,254 @@
...
+    MessagesForTagExpanded {
+        tag: String,
+        #[serde(rename = "maxResults")]
+        max_results: usize,
+        count: usize,
+        #[serde(rename = "messageIds")]
+        message_ids: Vec<Record>,
+    },
+    /// Response of GET /api/<keyspace>/addresses/<address>/outputs
```
bingyanglin, line 114:
```suggestion
    /// Response of GET /api/<api_version>/addresses/<address>/outputs
```

src/api/responses.rs
```diff
@@ -0,0 +1,254 @@
...
+    OutputsForAddress {
+        address: String,
+        #[serde(rename = "maxResults")]
+        max_results: usize,
+        count: usize,
+        #[serde(rename = "outputIds")]
+        output_ids: Vec<String>,
+    },
+    /// Response of GET /api/<keyspace>/addresses/<address>/outputs[?expanded=true]
```
bingyanglin, line 123:
```suggestion
    /// Response of GET /api/<api_version>/addresses/<address>/outputs[?expanded=true]
```

src/api/responses.rs
```diff
@@ -0,0 +1,254 @@
...
+    OutputsForAddressExpanded {
+        address: String,
+        #[serde(rename = "maxResults")]
+        max_results: usize,
+        count: usize,
+        #[serde(rename = "outputIds")]
+        output_ids: Vec<Record>,
+    },
+    /// Response of GET /api/<keyspace>/outputs/<output_id>
```
bingyanglin, line 132:
```suggestion
    /// Response of GET /api/<api_version>/outputs/<output_id>
```

src/api/responses.rs
```diff
@@ -0,0 +1,254 @@
...
+        message_id: String,
+        #[serde(rename = "transactionId")]
+        transaction_id: String,
+        #[serde(rename = "outputIndex")]
+        output_index: u16,
+        #[serde(rename = "spendingTransaction")]
+        spending_transaction: Option<Value>,
+        output: Value,
+    },
+    /// Response of GET /api/<keyspace>/transactions/<message_id>
```
bingyanglin, line 144:
```suggestion
    /// Response of GET /api/<api_version>/transactions/<message_id>
```

src/api/responses.rs
```diff
@@ -0,0 +1,254 @@
...
+    Transaction(Transaction),
+    /// Response of GET /api/<keyspace>/transactions/ed25519/<address>
```
bingyanglin, line 146:
```suggestion
    /// Response of GET /api/<api_version>/transactions/ed25519/<address>
```

src/api/responses.rs
```diff
@@ -0,0 +1,254 @@
...
+    Transactions { transactions: Vec<Transaction> },
+    TransactionHistory {
+        address: String,
+        transactions: Vec<Transfer>,
+    },
+    /// Response of GET /api/<keyspace>/milestone/<index>
```
bingyanglin, line 152:
```suggestion
    /// Response of GET /api/<api_version>/milestone/<index>
```

src/api/responses.rs
```diff
@@ -0,0 +1,254 @@
...
+    Milestone {
+        #[serde(rename = "index")]
+        milestone_index: u32,
+        #[serde(rename = "messageId")]
+        message_id: String,
+        timestamp: u64,
+    },
+    /// Response of GET /analytics/addresses[?start_timestamp=<u32>&end_timestamp=<u32>]
```
bingyanglin, line 160: Maybe this endpoint should also be consistent with other api calls? Like
`/api/<api_version>/analytics/addresses[?start_timestamp=<u32>&end_timestamp=<u32>]`.

src/types/mongo/mod.rs
```diff
@@ -0,0 +1,261 @@
...
+
+    fn get_as_string(&self, key: impl AsRef<str>) -> anyhow::Result<String>;
+
+    fn get_as_u8(&self, key: impl AsRef<str>) -> anyhow::Result<u8>;
+
+    fn get_as_u16(&self, key: impl AsRef<str>) -> anyhow::Result<u16>;
+
+    fn get_as_u32(&self, key: impl AsRef<str>) -> anyhow::Result<u32>;
+
+    fn get_as_u64(&self, key: impl AsRef<str>) -> anyhow::Result<u64>;
```
bingyanglin, line 205: Same as above.

src/types/mongo/shimmer.rs
```diff
@@ -0,0 +1,683 @@
...
+}
+
+pub fn message_from_doc(mut doc: Document) -> anyhow::Result<Message> {
+    let mut builder = MessageBuilder::new(Parents::new(
+        doc.take_array("parents")?
+            .into_iter()
+            .map(|p| message_id_from_bson(p))
+            .collect::<Result<Vec<_>, _>>()?,
+    )?)
+    .with_protocol_version(doc.get_i32("protocol_version")? as u8)
```
bingyanglin, line 112:
```suggestion
    // .with_protocol_version(doc.get_i32("protocol_version")? as u8)
```

src/types/mongo/cpt2.rs
```diff
@@ -0,0 +1,441 @@
...
+    );
+    doc.insert("merkle_proof", hex::encode(essence.merkle_proof()));
+    doc.insert("next_pow_score", essence.next_pow_score() as i64);
+    doc.insert(
+        "next_pow_score_milestone_index",
+        essence.next_pow_score_milestone_index() as i32,
+    );
+    doc.insert(
+        "public_keys",
+        essence.public_keys().iter().map(|p| hex::encode(p)).collect::<Vec<_>>(),
```
bingyanglin, line 267:
```suggestion
        essence.public_keys().iter().map(hex::encode).collect::<Vec<_>>(),
```

bingyanglin: Same as other places.

Cargo.toml
```diff
@@ -19,3 +19,39 @@ name = "inx-chronicle"
...
+bee-message-cpt2 = { git = "https://github.com/iotaledger/bee.git", branch = "mainnet-develop", default-features = false, features = ["serde"], package = "bee-message" }
+bee-rest-api-cpt2 = { git = "https://github.com/iotaledger/bee.git", branch = "mainnet-develop", default-features = false, package="bee-rest-api" }
+
+# Shimmer / Stardust types
+bee-message-shimmer = { git = "https://github.com/iotaledger/bee.git", branch = "shimmer-develop", default-features = false, features = ["std", "serde1", "dto"], package = "bee-message" }
+bee-rest-api-shimmer = { git = "https://github.com/iotaledger/bee.git", branch = "shimmer-develop", default-features = false, package="bee-rest-api" }
+
+[features]
+default = [ "api" ]
+api = [ "axum", "hyper", "tower", "tower-http" ]
```
grtlr, line 57:
```suggestion
api = [ "axum", "hyper", "tower", "tower-http" ]

```

src/api/mod.rs
```diff
@@ -0,0 +1,288 @@
...
+mod error;
+mod responses;
+mod routes;
+
+use std::{
+    task::Poll,
+    time::SystemTime,
+};
+
+use actix::prelude::*;
```
grtlr, line 13: Would it be possible to import the required stuff explicitly, or is it just too much? I'm not a big fan of preludes.

Cargo.toml
```diff
@@ -19,3 +19,39 @@ name = "inx-chronicle"
 path = "src/main.rs"
 
 [dependencies]
+actix = "0.13"
+anyhow = "1.0"
+cfg-if = "1.0"
+derive_more = "0.99"
+dotenv = "0.15"
+env_logger = "0.9"
```
grtlr, line 27: What do you think, should we maybe use `fern` instead?

.gitignore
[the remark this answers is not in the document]
grtlr: Should we also add the `.env` file?

Cargo.toml
```diff
@@ -19,3 +19,39 @@ name = "inx-chronicle"
 path = "src/main.rs"
 
 [dependencies]
+actix = "0.13"
+anyhow = "1.0"
+cfg-if = "1.0"
```
grtlr, line 24: I might be wrong, but do we actually use this right now?

src/main.rs
```diff
@@ -1,6 +1,32 @@
-// Copyright 2021 IOTA Stiftung
+// Copyright 2022 IOTA Stiftung
 // SPDX-License-Identifier: Apache-2.0
 
+use actix::prelude::*;
```
grtlr, line 4: Depending on other discussion, consider removing `prelude` here as well.

src/lib.rs
```diff
@@ -1,6 +1,32 @@
...
+pub mod cpt2 {
+    pub use bee_message_cpt2::{
+        self,
+        *,
+    };
+    pub use bee_rest_api_cpt2::{
+        self,
+        *,
+    };
+}
```
grtlr, line 18: If I remember correctly, we can just do `pub use bee_message_cpt2 as cpt2;`?

src/config/mongo.rs
```diff
@@ -0,0 +1,597 @@
...
+            retry_writes: Default::default(),
+            selection_criteria: Default::default(),
+            server_api: Default::default(),
+            server_selection_timeout: Default::default(),
+            default_database: Default::default(),
+            tls: Default::default(),
+            write_concern: Default::default(),
+        }
+    }
+}
```
grtlr, line 210: Maybe derive `Default`?

src/config/mongo.rs
```diff
@@ -0,0 +1,597 @@
...
+};
+use serde::{
+    Deserialize,
+    Serialize,
+};
+
+/// A clone of the `ClientOptions` structure from the `mongodb` crate
+/// which can be Serialized.
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
+pub struct MongoConfig {
```
grtlr, line 28: Should all of these options be configurable to the user? If not I think we should go with the builder pattern here. The builder can then be serializable and contains all the options that the user can specify (like address and such). In the `finish`/`build` step of the builder we can set the rest and return a `mongodb::ClientOptions` object.

If we follow this pattern for all major components, we can then aggregate all these builders to obtain our config file format.

src/types/mod.rs
```diff
@@ -0,0 +1,338 @@
...
+    Deserialize,
+    Serialize,
+};
+
+#[derive(Copy, Clone, Debug, Eq, PartialEq, Serialize, Deserialize, From, Hash, Ord, PartialOrd)]
+pub enum MessageId {
+    /// Chrysalis compatible message
+    Chrysalis(bee_message_cpt2::MessageId),
+    /// Shimmer compatible message
+    Shimmer(bee_message_shimmer::MessageId),
```
grtlr, line 28: Maybe we go with Stardust for now?

```suggestion
    Stardust(bee_message_shimmer::MessageId),
```

src/config/mongo.rs
```diff
@@ -0,0 +1,597 @@
...
+pub enum ServerAddress {
+    /// A TCP/IP host and port combination.
+    Tcp {
+        /// The hostname or IP address where the MongoDB server can be found.
+        host: String,
+
+        /// The TCP port that the MongoDB server is listening on.
+        ///
+        /// The default is 27017.
+        port: Option<u16>,
```
grtlr, line 256: Should we use something like `std::net::SocketAddr` here?

.gitignore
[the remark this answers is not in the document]
bingyanglin: I think yes.

src/api/responses.rs
grtlr: I think API version should be sufficient.

src/types/mongo/mod.rs
```diff
@@ -0,0 +1,261 @@
...
+
+    fn as_u16(&self) -> Result<u16, ValueAccessError>;
+
+    fn as_u32(&self) -> Result<u32, ValueAccessError>;
+
+    fn as_u64(&self) -> Result<u64, ValueAccessError>;
+
+    fn to_array(self) -> Result<Vec<Bson>, ValueAccessError>;
+
+    fn to_document(self) -> Result<Document, ValueAccessError>;
```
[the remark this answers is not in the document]
Alexandcoats: I don't like unnecessarily verbose function names and frankly this is all internal anyway.

Cargo.toml
```diff
@@ -19,3 +19,39 @@ name = "inx-chronicle"
...
+tower = { version = "0.4", optional = true }
+tower-http = { version = "0.2", features = ["cors", "catch-panic", "trace"], optional = true }
+
+# Chrysalis pt. 2 types
+bee-message-cpt2 = { git = "https://github.com/iotaledger/bee.git", branch = "mainnet-develop", default-features = false, features = ["serde"], package = "bee-message" }
+bee-rest-api-cpt2 = { git = "https://github.com/iotaledger/bee.git", branch = "mainnet-develop", default-features = false, package="bee-rest-api" }
+
+# Shimmer / Stardust types
+bee-message-shimmer = { git = "https://github.com/iotaledger/bee.git", branch = "shimmer-develop", default-features = false, features = ["std", "serde1", "dto"], package = "bee-message" }
+bee-rest-api-shimmer = { git = "https://github.com/iotaledger/bee.git", branch = "shimmer-develop", default-features = false, package="bee-rest-api" }
```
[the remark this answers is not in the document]
Alexandcoats: As long as we can find appropriate versions, then sure.

Cargo.toml
```diff
@@ -19,3 +19,39 @@ name = "inx-chronicle"
 path = "src/main.rs"
 
 [dependencies]
+actix = "0.13"
+anyhow = "1.0"
```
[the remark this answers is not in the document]
Alexandcoats: No, I just like anyhow.

rustfmt.toml
```diff
@@ -9,3 +9,4 @@ max_width = 120
 normalize_comments = true
 normalize_doc_attributes = true
 wrap_comments = true
+imports_layout = "Vertical"
```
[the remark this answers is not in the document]
Alexandcoats: I prefer it for readability reasons. If the imports list is too long that is a good sign that you should split up a file 🤷

src/config/mongo.rs
Alexandcoats: Check out the `hosts` line.

src/config/mongo.rs
Alexandcoats: The struct was originally designed to be read from a file, so I didn't bother with a builder, but yes it could use one. How do you feel about `derive-builder`?

src/types/mod.rs
Alexandcoats: 🤷

src/config/mongo.rs
Alexandcoats: No, because SocketAddr has validation that we do not necessarily want, and also forces a port to exist. However, we could create some helpful trait impls to allow parsing from strings and such similar to SocketAddr.

Cargo.toml
Alexandcoats: Yes, it is used for mongo config.

src/lib.rs
grtlr: Yeah, you're right, I probably didn't have enough coffee back then.

src/config/mongo.rs
grtlr: Ah, cool. If that's the case when can just keep it like this for now.

src/config/mongo.rs
Alexandcoats: Nevermind, it's not suitable for this case. I added some `with_` methods instead.

grtlr: Looking good! We are probably going to need some of these configuration options in the future, but to make reviewing simpler for everyone, I think I would only add the ones we actually use for now. When can then add additional parameters to config as we need them.

src/types/mod.rs
grtlr: First, there will be Stardust, then Shimmer will come later and might contain more changes to the message layouts.

src/main.rs
```diff
@@ -1,6 +1,30 @@
...
+    },
+};
+
 fn main() {
-    println!("Hello world from Chronicle!");
+    dotenv::dotenv().unwrap();
+    env_logger::init();
+    let system = System::new();
+    let mongo_config =
+        MongoConfig::default().with_credential(Credential::default().with_username("root").with_password("pass"));
```
grtlr, line 24: I think we should go with the `MongoConfig::new().with_xxx(...).finish()` paradigm here, because it gives us the flexibility to add additional logic in the `.finish()` method if needed (for example for validation).

I think the way we handle the configuration of the node in Bee is quite sensible: https://github.com/iotaledger/bee/blob/shimmer-develop/bee-node/bee-node/src/config.rs

rustfmt.toml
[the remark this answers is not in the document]
grtlr: The more I think about it the more I like it. 😀

src/main.rs
Alexandcoats: Yes, but the bee Node config is actually a builder pattern in the sense that all fields are required in the final struct. Our config struct only exists to mirror the `ClientOptions` struct from mongodb, because it cannot be Serialized (for some reason...), and it contains entirely optional fields. It makes more sense to use the `MongoConfig` struct itself as the builder and simply return the `ClientOptions` if requested in the `build/finish` step.

src/api/mod.rs
```diff
@@ -0,0 +1,294 @@
...
+    }
+}
+
+async fn shutdown_signal(recv: oneshot::Receiver<()>) {
+    if let Err(e) = recv.await {
+        log::error!("Error receiving shutdown signal: {}", e);
+    }
+}
+
+fn register_metrics() {
```
Alex6323, line 146: I would move the whole metrics stuff into its own `metrics` module.

src/api/routes.rs
```diff
@@ -0,0 +1,901 @@
...
+    database: Extension<Database>,
+    TimeRange {
+        start_timestamp,
+        end_timestamp,
+    }: TimeRange,
+) -> ListenerResult {
+    let start_milestone = start_milestone(&database, start_timestamp).await?;
+    let end_milestone = end_milestone(&database, end_timestamp).await?;
+
+    todo!()
```
Alex6323, line 896: Reminder

src/types/sync.rs
```diff
@@ -0,0 +1,26 @@
...
+
+#[derive(Debug, Clone, Default, Serialize)]
+pub struct SyncData {
+    /// The completed(synced and logged) milestones data
+    pub completed: Vec<Range<u32>>,
+    /// Synced milestones data but unlogged
+    pub synced_but_unlogged: Vec<Range<u32>>,
+    /// Gaps/missings milestones data
+    pub gaps: Vec<Range<u32>>,
+}
```
grtlr, line 26: I guess these two structs keep track of the amount of data that is available on the permanode? I would remove them from the PR and add an issue to the repository to remember that functionality?

rustfmt.toml
grtlr, line 12: Missing newline at the end of file:

```suggestion
imports_layout = "Vertical"

```

src/api/mod.rs
```diff
@@ -0,0 +1,294 @@
...
+    Layer,
+    Service,
+};
+
+use crate::config::mongo::MongoConfig;
+
+/// The Chronicle API actor
+#[derive(Debug, Serialize, Deserialize, Default)]
+pub struct ChronicleAPI {
+    mongo_config: MongoConfig,
```
grtlr, line 65: The API will probably be just one component of Chronicle. Should we store it as an `RwLock<Arc<MongoConfig>>` (or some other form of shared resource) here?

src/api/mod.rs
```diff
@@ -0,0 +1,294 @@
...
+
+    fn started(&mut self, _ctx: &mut Self::Context) {
+        register_metrics();
+        let client_opts: ClientOptions = self.mongo_config.clone().into();
+        log::info!("Connecting to MongoDB");
+        let client = Client::with_options(client_opts)
+            .map_err(|e| {
+                log::error!("Unable to create client from options, error: {}", e);
+            })
+            .unwrap();
```
grtlr, line 105: Move these over to the construction of the `ChronicleAPI` so that we can avoid the `.unwrap()` calls. For example like: `ChronicleAPI::new() -> Result<_,_>`.

src/api/mod.rs
```diff
@@ -0,0 +1,294 @@
...
+        let db = client.database("permanode");
+        let (sender, receiver) = oneshot::channel();
+        log::info!("Starting Axum server");
+        tokio::spawn(async {
+            Server::bind(&([0, 0, 0, 0], 8000).into())
```
grtlr, line 110: Make configurable through CLI / configuration file.

src/api/mod.rs
grtlr: Maybe we can even leave out the metrics for now and re-add them at a later point?

src/api/responses.rs
```diff
@@ -0,0 +1,220 @@
...
+
+use super::extractors::{
+    MessagesQuery,
+    OutputsQuery,
+};
+use crate::types::LedgerInclusionState;
+
+#[derive(Clone, Debug, Serialize, Deserialize)]
+#[serde(untagged)]
+pub(crate) enum ListenerResponse {
```
grtlr, line 22: Do we need all these responses in the same `enum`, or could we just use `impl IntoResponse` where appropriate?

src/api/mod.rs
Alexandcoats: If the metrics aren't going to be used internally, I doubt they need to exist.

rustfmt.toml
Alex6323: Can we not add this format rule?

src/types/mongo/stardust.rs
```diff
@@ -0,0 +1,668 @@
...
+    let mut doc = Document::new();
+    doc.insert("protocol_version", message.protocol_version() as i32);
+    doc.insert(
+        "parents",
+        message.parents().iter().map(|p| p.to_string()).collect::<Vec<_>>(),
+    );
+    doc.insert("payload", message.payload().map(payload_to_bson));
+    doc.insert("nonce", message.nonce().to_string());
+    Bson::Document(doc)
+}
```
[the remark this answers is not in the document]
Alexandcoats: There are several problems with storing the message directly.
1. Bson cannot store u64, at all, ever. If we have u64 data, it will try to convert to an i64 and fail if it is out of range. This is only important for the nonce as far as I can tell, but it needs to be explicitly de/serialized as a string.
Now, if that is the only u64 that can possibly be out of range, then we can simply manually handle the top level and use the serde impl for the payload.
2. In order to create consistent indexes on the data, we need to add fields to the stored document in strategic places. For instance, several of the `index-able` fields are inside lists of blocks in stardust, but some, like the address, are fundamentally the same index as cpt2. Therefore, the simplest solution to index both is to create a field in the same place on both documents. This has to be at the output level conceptually, so it can't just be at the root. Similarly, timestamps must be converted to Bson `Dates` to be useful, versus directly serializing them as i64.
3. The `bee` serde impls use externally tagged enums, meaning that all levels are effectively two levels (i.e. `payload.essence.outputs.address` vs `payload.data.essence.data.outputs.data.address.data`) which makes matching and pipelines much more difficult. If we're going to use their impls, I think it would be in our best interest to change that to internal tags.

I don't think it really makes sense to change the protocol to support Bson. Ideally this data could be stored anywhere, but I think it's reasonable to expect that some internal conversions must take place. I think we could definitely reduce this and the effort needed for each protocol release by serializing the payloads and adding fields to the resulting document as needed, but there are good reasons why I did it this way for the time being.

src/types/sync.rs
Alexandcoats: They are fundamental to the design, as we cannot know when to request milestones unless we know which ones we have already gathered every message for.

src/api/mod.rs
Alexandcoats: Should probably leave this for another PR, yes?

src/api/responses.rs
Alexandcoats: There are pros and cons to both approaches, but either work.

src/api/mod.rs
grtlr: I'd say we remove them for now, and them back later using Prometheus. What do you think?

Alexandcoats: Well they are already using prometheus, FYI, and are fully functional.

src/types/mod.rs
```diff
@@ -0,0 +1,79 @@
...
+}
+
+impl TryFrom<u8> for LedgerInclusionState {
+    type Error = anyhow::Error;
+    fn try_from(value: u8) -> Result<Self, Self::Error> {
+        match value {
+            0 => Ok(Self::Conflicting),
+            1 => Ok(Self::Included),
+            2 => Ok(Self::NoTransaction),
+            n => bail!("Unexpected ledger inclusion byte state: {}", n),
```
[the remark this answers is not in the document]
Alexandcoats: Well, it's certainly not infallable, or it would be a `From` impl. Is that error not meaningful?

src/api/mod.rs
grtlr: Yeah, sorry. I just realized that.

src/api/routes.rs
```diff
@@ -0,0 +1,876 @@
...
+                .sort(doc! {"milestone_index": 1})
+                .limit(1)
+                .build(),
+        )
+        .await?
+        .try_next()
+        .await?
+        .map(|mut d| {
+            d.get_document_mut("message")
+                .unwrap()
```
pvdrz, line 301: undocumented `unwrap`s

src/api/routes.rs
```diff
@@ -0,0 +1,876 @@
...
+            },
+            None,
+        )
+        .await?;
+
+    Ok(ListenerResponse::Output {
+        message_id: output.get_str("message_id").unwrap().to_owned(),
+        transaction_id: transaction_id.to_string(),
+        output_index: idx,
+        spending_transaction: spending_transaction.map(|mut d| d.remove("message").unwrap().into()),
```
pvdrz, line 437: undocumented `unwrap`

src/api/routes.rs
```diff
@@ -0,0 +1,876 @@
...
+        output: output
+            .get_document_mut("message")
+            .unwrap()
```
pvdrz, line 440: same

src/api/routes.rs
```diff
@@ -0,0 +1,876 @@
...
+        .await?;
+
+    let transactions = records
+        .into_iter()
+        .map(|rec| {
+            let payload = rec.get_document("message").unwrap().get_document("payload").unwrap();
+            let spending_transaction = rec.get_document("spending_transaction").ok();
+            let output = payload
+                .get_document("essence")
+                .unwrap()
```
pvdrz, line 634: same

src/api/routes.rs
```diff
@@ -0,0 +1,876 @@
...
+        message_id,
+        milestone_index: transaction.milestone_index,
+        outputs: match &transaction.message {
+            Message::Chrysalis(m) => match m.payload() {
+                Some(crate::cpt2::payload::Payload::Transaction(t)) => match t.essence() {
+                    crate::cpt2::prelude::Essence::Regular(e) => {
+                        e.outputs().iter().map(|o| serde_json::to_value(o).unwrap()).collect()
+                    }
+                },
+                _ => unreachable!(),
```
pvdrz, line 676: undocumented `panic`

src/api/routes.rs
```diff
@@ -0,0 +1,876 @@
...
+            },
+            Message::Stardust(m) => match m.payload() {
+                Some(crate::stardust::payload::Payload::Transaction(t)) => match t.essence() {
+                    crate::stardust::payload::transaction::TransactionEssence::Regular(e) => {
+                        e.outputs().iter().map(|o| serde_json::to_value(o).unwrap()).collect()
+                    }
+                },
+                _ => unreachable!(),
```
pvdrz, line 684: same

src/api/routes.rs
```diff
@@ -0,0 +1,876 @@
...
+            },
+        },
+        inputs: match &transaction.message {
+            Message::Chrysalis(m) => match m.payload() {
+                Some(crate::cpt2::payload::Payload::Transaction(t)) => match t.essence() {
+                    crate::cpt2::prelude::Essence::Regular(e) => {
+                        e.inputs().iter().map(|o| serde_json::to_value(o).unwrap()).collect()
+                    }
+                },
+                _ => unreachable!(),
```
pvdrz, line 694: same

src/api/routes.rs
```diff
@@ -0,0 +1,876 @@
...
+            },
+            Message::Stardust(m) => match m.payload() {
+                Some(crate::stardust::payload::Payload::Transaction(t)) => match t.essence() {
+                    crate::stardust::payload::transaction::TransactionEssence::Regular(e) => {
+                        e.inputs().iter().map(|o| serde_json::to_value(o).unwrap()).collect()
+                    }
+                },
+                _ => unreachable!(),
```
pvdrz, line 702: same

src/api/routes.rs
```diff
@@ -0,0 +1,876 @@
...
+                            m.essence().timestamp()
+                        } else {
+                            unreachable!()
+                        }
+                    }
+                    Message::Stardust(m) => {
+                        if let Some(crate::stardust::payload::Payload::Milestone(m)) = m.payload() {
+                            m.essence().timestamp()
+                        } else {
+                            unreachable!()
```
pvdrz, line 769: same

src/api/routes.rs
```diff
@@ -0,0 +1,876 @@
...
+) -> ListenerResult {
+    let start_milestone = start_milestone(&database, start_timestamp).await?;
+    let end_milestone = end_milestone(&database, end_timestamp).await?;
+
+    let res = database
+        .collection::<Document>("messages")
+        .aggregate(
+            vec![
+                doc! { "$match": {
+                    "inclusion_state": LedgerInclusionState::Included as u8 as i32,
```
pvdrz, line 792: I think you can do `.into()` instead of `as i32`

src/api/routes.rs
```diff
@@ -0,0 +1,876 @@
...
+                    "milestone_index": { "$gt": start_milestone, "$lt": end_milestone },
+                    "message.payload.kind": crate::cpt2::payload::transaction::TransactionPayload::KIND as i32,
```
pvdrz, line 794: same

src/api/routes.rs
```diff
@@ -0,0 +1,876 @@
...
+                } },
+                doc! { "$unwind": { "path": "$message.payload.essence.inputs", "includeArrayIndex": "message.payload.essence.inputs.idx" } },
+                doc! { "$lookup": {
+                    "from": "messages",
+                    "let": { "transaction_id": "$message.payload.essence.inputs.transaction_id", "index": "$message.payload.essence.inputs.index" },
+                    "pipeline": [
+                        { "$match": { 
+                            "inclusion_state": LedgerInclusionState::Included as u8 as i32,
```
pvdrz, line 802: same

src/api/routes.rs
```diff
@@ -0,0 +1,876 @@
...
+                    "recv_addresses": { "$arrayElemAt": ["$recv.addresses", 0] },
+                    "send_addresses": { "$arrayElemAt": ["$send.addresses", 0] },
+                } },
+            ],
+            None,
+        )
+        .await?.try_next().await?.ok_or_else(|| anyhow::anyhow!("No transactions found in time range"))?;
+
+    Ok(ListenerResponse::AddressAnalytics {
+        total_addresses: res.get("total_addresses").unwrap().as_u64().unwrap(),
```
pvdrz, line 852: undocumented `unwrap`s

src/config/mongo.rs
```diff
@@ -0,0 +1,884 @@
...
+use serde::{
+    Deserialize,
+    Serialize,
+};
+
+/// A clone of the [`ClientOptions`] structure from the [`mongodb`] crate
+/// which can be Serialized.
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
+pub struct MongoConfig {
+    /// The initial list of seeds that the Client should connect to.
```
pvdrz, line 29: :heart: docs

src/config/mongo.rs
```diff
@@ -0,0 +1,884 @@
...
+    pub tls: Option<Tls>,
+
+    /// Specifies the default write concern for operations performed on the Client. See the
+    /// WriteConcern type documentation for more details.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub write_concern: Option<WriteConcern>,
+}
+
+#[allow(missing_docs)]
+impl MongoConfig {
```
pvdrz, line 184: I wonder if we should use some builder crate in the future

src/main.rs
```diff
@@ -1,10 +1,36 @@
...
+    let res: anyhow::Result<()> = system.block_on(async {
+        let api_addr = ChronicleAPI::new(mongo_config)?.start();
+        tokio::signal::ctrl_c().await.ok();
```
pvdrz, line 31: why do we ignore these errors?

src/types/mongo/cpt2.rs
```diff
@@ -0,0 +1,436 @@
...
+        IndexationPayload::KIND => Payload::Indexation(Box::new(indexation_payload_from_bson(bson)?)),
+        ReceiptPayload::KIND => Payload::Receipt(Box::new(receipt_payload_from_bson(bson)?)),
+        TreasuryTransactionPayload::KIND => Payload::TreasuryTransaction(Box::new(treasury_payload_from_bson(bson)?)),
+        _ => bail!("Unknown payload kind: {}", kind),
+    })
+}
+
+fn transaction_payload_to_bson(payload: &TransactionPayload) -> Bson {
+    let mut doc = Document::new();
+    doc.insert("kind", TransactionPayload::KIND as i32);
```
pvdrz, line 88: use `.into()` if possible.

pvdrz (changes requested):

src/main.rs
Alexandcoats: Laziness?

src/config/mongo.rs
Alexandcoats: I considered using `derive-builder`, but the struct is basically a builder already, so it didn't seem necessary to complicate the access and add another type.

src/api/routes.rs
Alexandcoats: Try it and you'll see why I can't

src/main.rs
pvdrz: :trollface: but can't they be coerced into `anyhow::Error`?

src/api/error.rs
```diff
@@ -0,0 +1,90 @@
...
+
+impl ListenerError {
+    pub fn status(&self) -> StatusCode {
+        match self {
+            ListenerError::NoResults | ListenerError::NotFound => StatusCode::NOT_FOUND,
+            ListenerError::IndexTooLarge
+            | ListenerError::InvalidHex
+            | ListenerError::BadParse(_)
+            | ListenerError::QueryError(_) => StatusCode::BAD_REQUEST,
+            _ => StatusCode::INTERNAL_SERVER_ERROR,
```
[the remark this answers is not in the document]
Alexandcoats: Fair point

src/types/mod.rs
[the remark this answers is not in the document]
grtlr: I think I phrased it wrong, I meant an actual `Error` type.

src/api/metrics.rs
```diff
@@ -0,0 +1,210 @@
...
+        let this = self.project();
+        match this.fut.poll(cx) {
+            Poll::Ready(res) => {
+                let duration = this.start_time.elapsed().unwrap();
+                let ms = (duration.as_secs() * 1000 + duration.subsec_millis() as u64) as f64;
+                RESPONSE_TIME_COLLECTOR
+                    .with_label_values(&[&format!("{} {}", this.method, this.uri)])
+                    .observe(ms);
+                if let Ok(res) = res.as_ref() {
+                    match res.status().as_u16() {
```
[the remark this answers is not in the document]
Alexandcoats: Probably don't want to use modulo there 😉

src/types/mongo/cpt2.rs
Alexandcoats: Coercing u32 to i32 is fallible operation, so that would look like this:

```
doc.insert("kind", TryInto::<i32>::try_into(TransactionPayload::KIND).unwrap());
```

🤷

PR #5, opened 2022-03-31
Deny `warnings` and `missing-docs` in CI

grtlr:
I think we should deny warnings and documentation in CI. In general, I think these are good defaults, and it is more convenient for development to do that in CI, rather than adding `#[deny(...)]` to the crates. What do you think?


bingyanglin (approved):

Alexandcoats:
If we're going to do that we should use `#![warn(missing_docs)]` so that we remember to do it before creating a PR.

PR #26, opened 2022-04-06
Main inx backstage

Alexandcoats:
This is a version of the INX hello world test that uses backstage, and shows how it can be used to create a supervision tree which can flexibly handle child exit requests.

## Pros
- Supervision guarantees (which extend to non-actor tasks spawned with the context)
- Central resource store makes sharing things like the database simple
- Async traits and error handling built in

## Cons
- Immature syntax and features in backstage mean that some things are awkward

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,7 +19,8 @@ name = "inx-chronicle"
 path = "src/main.rs"
 
 [dependencies]
+axum = { version = "0.5.0", default-features = false, features = [ "json", "headers" ] }
 futures = { version = "0.3.21", default-features = false }
 inx = { git = "https://github.com/iotaledger/inx", version = "0.1.0", default-features = false }
-tokio = { version = "1.17.0", default-features = false, features = ["macros", "rt-multi-thread"] }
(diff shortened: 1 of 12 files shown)


Cargo.toml
```diff
@@ -1,13 +1,13 @@
 [package]
 name = "chronicle-inx"
 version = "0.1.0"
-authors = [ "IOTA Stiftung" ]
+authors = ["IOTA Stiftung"]
 edition = "2021"
-description = "IOTA permanode implemented as an IOTA Node Extension (INX)"
+description = "IOTA permanode implemented as an IOTA Node Extension (INX)."
```
Alex6323, line 6: punctuation :heart:

Cargo.toml
```diff
@@ -1,13 +1,13 @@
...
 readme = "README.md"
 repository = "https://github.com/iotaledger/inx-chronicle"
 license = "Apache-2.0"
-keywords = [ "iota", "storage", "permanode", "node", "inx" ]
+keywords = ["iota", "storage", "permanode", "node", "inx"]
```
Alex6323, line 10: "permanode" and "node" are a bit redundant. I'ld replace "node" for "chronicle" as this is the brand name.

PR #27, opened 2022-04-06
Main inx basic

Alexandcoats:
No frameworks.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -3,8 +3,29 @@
 
 #![warn(missing_docs)]
 
-//! TODO
+use futures::stream::StreamExt;
+use inx::{client::InxClient, proto::NoParams};
+use tonic::transport::Channel;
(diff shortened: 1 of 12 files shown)


src/main.rs
```diff
@@ -5,6 +5,288 @@
...
+                        error!("Error while listening to milestones: {}", err);
+                        if self.handle.send(INXListenerEvent::Shutdown).is_err() {
+                            break;
+                        }
+                    }
+                }
+                INXListenerEvent::Shutdown => break,
+            }
+        }
+        // This isn't really good enough, due to early exits :(
```
pvdrz, line 174: We could fix this by creating a wrapper around `AbortHandle` and implementing `Drop` for it so it calls `handle.abort()`. What do you think?

pvdrz (approved):

PR #28, opened 2022-04-07
Rudimentary INX and plumbing

grtlr:


Cargo.toml
```diff
@@ -1,21 +1,25 @@
...
-
-[[bin]]
-name = "inx-chronicle"
-path = "src/main.rs"
+[workspace]
+default-members = [".", "bin/inx-chronicle"]
+members = [
+    ".",
+    "bin/inx-chronicle",
+]
```
Alex6323, line 18: Can you explain to me why `default-members` and `members` are the same. I am not too familiar with those fields.

.gitignore
```diff
@@ -1 +1,9 @@
 /target
+
+.env
+
+# Hornet files
+/alphanet
+config.json
+peering.json
+profiles.json
```
Alex6323, line 9: Why are those excluded here? :thinking: Shouldn't inx-chronicle be independent of the concrete node impl, or is this just temporary?

Cargo.toml
```diff
@@ -1,21 +1,28 @@
...
-authors = [ "IOTA Stiftung" ]
+authors = ["IOTA Stiftung"]
 edition = "2021"
-description = "IOTA permanode implemented as an IOTA Node Extension (INX)"
+description = "IOTA permanode implemented as an IOTA Node Extension (INX)."
 readme = "README.md"
 repository = "https://github.com/iotaledger/inx-chronicle"
 license = "Apache-2.0"
-keywords = [ "iota", "storage", "permanode", "node", "inx" ]
+keywords = ["iota", "storage", "permanode", "node", "inx"]
```
Alex6323, line 10: I would replace "node" with "chronicle".

README.md
```diff
@@ -1 +1,13 @@
...
+
+The easiest way to get going is to use a `private_tangle` for now:
+```sh
+git clone git@github.com:gohornet/hornet.git
+cd hornet/private_tangle
+./bootstrap.sh # exit with <CTRL-C>
+./run.sh
+```
+
+Then you should be able to connect to INX on `http://localhost:9029`
```
Alex6323, line 13: cool, it's really easy :heart:

Cargo.toml
grtlr: I was just experimenting. 😄

src/lib.rs
```diff
@@ -1,14 +1,11 @@
...
 //! TODO
 
-#[cfg(test)]
-mod test {
-    #[test]
-    fn test_fn() {
-        assert_eq!(true, true);
-    }
-}
+/// Module that contains the database types and constans.
```
Alex6323, line 8:
```suggestion
/// Module that contains the database types and constants.
```

src/inx/mod.rs
```diff
@@ -0,0 +1,29 @@
...
+
+/// A builder to establish a connection to INX.
+#[must_use]
+#[derive(PartialEq, Eq, Debug, Serialize, Deserialize)]
+pub struct InxConfig {
+    address: String,
+}
+
+impl InxConfig {
+    /// Creates a new [`InxConfig`]. The `address` is the address of the nodes INX interface.
```
Alex6323, line 18:
```suggestion
    /// Creates a new [`InxConfig`]. The `address` is the address of the node's INX interface.
```

src/inx/mod.rs
```diff
@@ -0,0 +1,29 @@
...
+mod error;
+use inx::{client::InxClient, Channel};
+use serde::{Deserialize, Serialize};
+
+pub use self::error::InxError;
+
+/// A builder to establish a connection to INX.
+#[must_use]
+#[derive(PartialEq, Eq, Debug, Serialize, Deserialize)]
+pub struct InxConfig {
```
Alex6323, line 13: That's not a builder anymore, but the actual config? If so you can remove #[must_use] and adapt the doc.

src/inx/mod.rs
```diff
@@ -0,0 +1,29 @@
...
+}
+
+impl InxConfig {
+    /// Creates a new [`InxConfig`]. The `address` is the address of the nodes INX interface.
+    pub fn new(address: String) -> Self {
+        Self { address }
+    }
+
+    /// Constructs an [`InxClient`] by consuming the [`InxConfig`].
+    pub async fn build(self) -> Result<InxClient<Channel>, InxError> {
```
Alex6323, line 24: It's probably easier to understand if there's a static `build` function on `InxClient` instead (which consumes the config), so like this:
```rust
impl InxClient<Channel> {
    pub async fn build(inx_config: InxConfig) -> Result<Self, InxError> {...}
}
But that's something I can also change in another PR if you want.

File: src/db/mod.rs [hunk 9]
```diff
@@ -0,0 +1,114 @@
...
+};
+use serde::{Deserialize, Serialize};
+
+/// Name of the MongoDB database.
+pub const DB_NAME: &str = "chronicle-test";
+
+/// A builder to establish a connection to the database.
+#[must_use]
+#[derive(PartialEq, Eq, Debug, Serialize, Deserialize)]
+pub struct MongoConfig {
```
Alex6323, line 23: That's a bit different to what we did in Bee, but it seems to be on purpose, right?

What about the pattern `let mongo_db = MongoConfig::build().with_a().with_b().finish();`?

I really like that pattern. Or is this just temporary and supposed to be changed by me in the config PR?

grtlr: Yeah, I figured these will change anyway so I didn't want to bother separating `{Inx|MongoDb}Config` and `{Inx|MongoDb}ConfigBuilder`. Would that be ok, for you?

src/inx/mod.rs
Alex6323: I would also not call it `build`, but `init`.

pvdrz (approved):

Alexandcoats (approved):

PR #29, opened 2022-04-07
Add simplified actor runtime

Alexandcoats:
A.K.A. `backstage-lite`


src/runtime/actor/envelope.rs
```diff
@@ -0,0 +1,50 @@
...
+pub trait DynEvent<A: Actor>: Debug {
+    fn handle<'c, 'a>(
+        self: Box<Self>,
+        cx: &'c mut ActorContext<A>,
+        act: &'c mut A,
+        data: &'c mut A::Data,
+    ) -> Pin<Box<dyn core::future::Future<Output = Result<(), A::Error>> + Send + 'a>>
+    where
+        Self: 'a,
+        'c: 'a;
```
grtlr, line 29: Do we need the `'c` lifetime here, if it's the same as `'a`?

src/runtime/actor/context.rs
```diff
@@ -0,0 +1,76 @@
...
+    fn deref(&self) -> &Self::Target {
+        &self.scope
+    }
+}
+
+impl<A: Actor> DerefMut for ActorContext<A> {
+    fn deref_mut(&mut self) -> &mut Self::Target {
+        &mut self.scope
+    }
+}
```
grtlr, line 76: I'm not sure about the `Deref` here, I feel like it makes it harder to understand what's going on when accessing `cx.resource<>` in the implementation of an `Actor`. I don't think `cx.scope.resource<>` is too bad, but that may be personal preference.

src/runtime/actor/envelope.rs
```diff
@@ -0,0 +1,50 @@
...
+pub trait HandleEvent<E: Send + Debug>: Actor + Sized {
+    async fn handle_event(
+        &mut self,
+        cx: &mut ActorContext<Self>,
+        event: E,
+        data: &mut Self::Data,
+    ) -> Result<(), Self::Error>;
+}
+
+pub trait DynEvent<A: Actor>: Debug {
```
grtlr, line 20: Could you maybe explain in the documentation what's going on here? What is the purpose of `DynEvent`?

bin/inx-chronicle/config.toml
```diff
@@ -1,5 +1,7 @@
 [mongodb]
 location = "mongodb://localhost:27017"
+username = "root"
+password = "pass"
```
grtlr, line 4: We shouldn't supply configs that don't work with MongoDB out of the box. Commenting these out would still signal the user the presence of these fields.

From a developers perspective, it might make sense to also look in `dotenv` for these values, that way we wouldn't have to ping pong back and forth because of different individual configurations.

```suggestion
# username = "root"
# password = "pass"
```

bin/inx-chronicle/src/main.rs
```diff
@@ -8,35 +8,140 @@ mod cli;
...
+        scope.add_resource(db).await;
     } else {
-        (
-            MongoConfig::new("mongodb://localhost:27017".into()).into_db().await?,
-            InxClient::connect("http://localhost:9029").await?,
-        )
-    };
+        let db = MongoConfig::new("mongodb://localhost:27017".into())
+            .with_username("root")
+            .with_password("pass")
```
grtlr, line 136: Same as above.
```suggestion
```

grtlr (changes requested):
This is looking really great! A couple of minor issues that I noted:

* [x] I can't use <kbd>Ctrl</kbd> + <kbd>C</kbd> to exit after an `ERROR` occured.
* [ ] When we can't connect to INX, the error message is very noisy because of the nesting of `Actor`s: `[2022-04-08T07:03:45Z ERROR inx_chronicle] Error launching scope: Actor exited with error: Actor exited with error: tonic transport failed`
* [ ] Warning from `anymap` about `clone_to_any_send`

I have also opened a PR against your branch with some minor changes and improvements: https://github.com/Alexandcoats/inx-chronicle/pull/3

src/runtime/actor/mod.rs
```diff
@@ -0,0 +1,60 @@
...
+
+#[async_trait]
+pub trait Actor: Send + Sync + Sized {
+    type Data: Send + Sync;
+    type Error: Error + Send + Sync;
+
+    /// Set this actor's name, primarily for debugging purposes
+    fn name(&self) -> String {
+        std::any::type_name::<Self>().into()
+    }
```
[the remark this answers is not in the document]
Alexandcoats: I used a `Cow` in backstage, but wasn't sure there was any point. Would you prefer that?

src/runtime/actor/mod.rs
```diff
@@ -0,0 +1,60 @@
...
+
+    /// Synchronously initialize this actor
+    async fn init(&mut self, cx: &mut ActorContext<Self>) -> Result<Self::Data, Self::Error>
+    where
+        Self: 'static + Sized + Send + Sync;
```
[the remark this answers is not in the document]
Alexandcoats: Yeah, overlooked them. Thanks.

src/runtime/actor/envelope.rs
Alexandcoats: Egh. This in particular took an incredible amount of tinkering to get working. Maybe there's a better way to express this but I couldn't find it.

src/runtime/actor/envelope.rs
Alexandcoats: This is what enables sending any event type to an actor over the same channel as long as they have implemented `HandleEvent` for it.

bin/inx-chronicle/config.toml
Alexandcoats: Sure, but not having any username and password will also fail so 🤷

src/runtime/actor/envelope.rs
grtlr: When I removed the `'c` it still compiled, so it must be working :trollface:

bin/inx-chronicle/src/listener.rs
```diff
@@ -4,72 +4,176 @@
...
-    let mut stream = response.unwrap().into_inner();
+#[async_trait]
+impl Actor for InxStreamListener<inx::proto::Message> {
+    type Data = (InxClient<Channel>, Act<Broker>);
+    type Error = INXListenerError;
+
+    async fn init(&mut self, cx: &mut ActorContext<Self>) -> Result<Self::Data, Self::Error>
+    where
+        Self: 'static + Sized + Send + Sync,
+    {
```
grtlr, line 123: These can also be removed for all actors.

src/runtime/actor/envelope.rs
Alexandcoats: Apparently it's fine. Weird.

src/runtime/scope.rs
```diff
@@ -0,0 +1,439 @@
...
+};
+use crate::runtime::{
+    actor::{envelope::Envelope, error::ActorError, report::ErrorReport},
+    shutdown::ShutdownStream,
+};
+
+/// A view into a particular scope which provides the user-facing API
+#[derive(Clone, Debug)]
+#[repr(transparent)]
+pub struct ScopeView(pub(crate) Scope);
```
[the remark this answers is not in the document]
Alexandcoats: Mostly due to naming of methods, since there would be a lot of collisions. Could always rename those to something like `fn internal_x()`. 🤷

bin/inx-chronicle/config.example.toml
```diff
@@ -0,0 +1,7 @@
+[mongodb]
+location = "mongodb://localhost:27017"
+username = "root"
+password = "root"
+
+[inx]
+address = "http://116.203.35.39:9029"
```
Alex6323, line 7: That's impossible! (officially :trollface:)

bin/inx-chronicle/config.toml
```diff
@@ -1,5 +1,7 @@
 [mongodb]
 location = "mongodb://localhost:27017"
+username = "root"
+password = "root"
 
 [inx]
-address = "http://localhost:9029"
```
Alex6323, line 5: :trollface:

src/runtime/actor/mod.rs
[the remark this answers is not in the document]
grtlr: What would be the advantage of using `Cow` over `&str`?

Alexandcoats:
Closing this for now. We may want to reference it in the future to re-implement shared resources after #33 is merged.

PR #31, opened 2022-04-12
Start implementing the data model

grtlr:
⚠️ Superseded by #55.


Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "chronicle"
-version = "0.1.0"
```
Alex6323, line 3: why that?

grtlr: I want to pre-release some Docker images to the GitHub container registry and I want to make sure that no one actually tries to use them 😁

Alexandcoats: Does this really determine the docker version?

Cargo.toml
```diff
@@ -19,13 +19,15 @@ name = "inx-chronicle"
 path = "bin/inx-chronicle/src/main.rs"
 
 [dependencies]
+bee-message-stardust = { package = "bee-message", version = "0.2", git = "https://github.com/iotaledger/bee", branch = "shimmer-develop", default-features = false, features = ["std", "serde1", "dto"], optional = true }
```
Alexandcoats, line 22: Separate optional dependencies?

Cargo.toml
```diff
@@ -19,13 +19,15 @@ name = "inx-chronicle"
...
+
 async-recursion = { version = "1.0", default-features = false }
 async-trait = { version = "0.1", default-features = false }
-clap = { version = "3.1", default-features = false, features = [ "derive", "std" ] }
+clap = { version = "3.1", default-features = false, features = ["derive", "std"] }
 dotenv = { version = "0.15", default-features = false }
 env_logger = { version = "0.9", default-features = false, features = ["termcolor", "atty", "humantime"] }
 futures = { version = "0.3", default-features = false }
-inx = { git = "https://github.com/iotaledger/inx", version = "0.2", default-features = false }
+inx = { git = "https://github.com/grtlr/inx", branch = "types", version = "0.2.0", default-features = false, features = ["types"] }
```
Alexandcoats, line 30: Use a patch instead?

src/db/model/mod.rs
```diff
@@ -0,0 +1,15 @@
+// Copyright 2022 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+#[cfg(feature = "stardust")]
+/// Models for Stardust layouts.
+pub mod stardust;
+
+use serde::Serialize;
+
+// TODO: Add `serde::Deserialize` constraint.
```
Alexandcoats, line 10: Probably just use `DeserializeOwned`.

src/db/model/stardust.rs
```diff
@@ -0,0 +1,39 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+use bee_message_stardust as stardust;
+use serde::{Deserialize, Serialize};
+
+use super::Model;
+
+/// Model for the [`Message`].
+#[derive(Serialize, Deserialize)]
+pub struct Message {
```
Alexandcoats, line 11: We should probably go ahead and include optional metadata.

src/db/model/stardust.rs
```diff
@@ -0,0 +1,39 @@
...
+    const COLLECTION: &'static str = "stardust_messages";
+}
+
+/// Model
+#[derive(Serialize, Deserialize)]
+pub struct Milestone {
+    /// The milestone index.
+    pub milestone_index: u32,
+    /// The timestamp of the milestone.
+    pub milestone_timestamp: u32,
```
Alexandcoats, line 30: Can we make this a `bson::DateTime`?

bin/inx-chronicle/src/main.rs
```diff
@@ -104,6 +104,7 @@ impl HandleEvent<Report<Broker>> for Launcher {
                                 cx.shutdown();
                             }
                         },
+                        other => log::warn!("Unhandled MongoDB error: {other}"),
```
Alexandcoats, line 107: Should call `cx.shutdown()` here, since we're not handling the error.

Cargo.toml
grtlr: I will create tags on GitHub that will build the Docker image automatically (with the tags as semver)

src/db/model/stardust.rs
grtlr: All of these: https://github.com/iotaledger/inx/blob/develop/proto/inx.proto#L128-L153?

Alexandcoats: Yeah, except maybe parents because that's on the message too.

Cargo.toml
```diff
@@ -21,16 +21,25 @@ path = "bin/inx-chronicle/src/main.rs"
...
-clap = { version = "3.1", default-features = false, features = [ "derive", "std" ] }
+clap = { version = "3.1", default-features = false, features = ["derive", "std"] }
 dotenv = { version = "0.15", default-features = false }
 env_logger = { version = "0.9", default-features = false, features = ["termcolor", "atty", "humantime"] }
 futures = { version = "0.3", default-features = false }
-inx = { git = "https://github.com/iotaledger/inx", version = "0.2", default-features = false }
+inx = { path = "../inx/rust", version = "0.2", default-features = false, features = ["types"] }
 log = { version = "0.4", default-features = false }
-mongodb = { version = "2.1", default-features = false, features = ["tokio-runtime"] }
+mongodb = { version = "2.1", default-features = false, features = ["bson-chrono-0_4", "tokio-runtime"] }
```
Alexandcoats, line 30: The `bson-chrono-0_4` feature should not be needed here.

grtlr: You're right, thanks!

PR #32, opened 2022-04-12
Add custom u64 serializer

pvdrz:


src/bson.rs
```diff
@@ -0,0 +1,461 @@
...
+    ) -> Result<Self::SerializeStructVariant, Self::Error> {
+        Ok(StructVariantSerializer {
+            name: variant,
+            inner: Document::new(),
+        })
+    }
+}
+
+struct ArraySerializer {
+    inner: Vec<Bson>,
```
Alexandcoats, line 264: Just for consistency, should this be an `Array` too?

Alexandcoats (changes requested):

src/bson.rs
pvdrz: The original code was like that. I don't think there is any issue with using `Array` tho

Alexandcoats (approved):

grtlr (approved):

PR #33, opened 2022-04-12
Runtime without shared resources

Alexandcoats:


bin/inx-chronicle/config.toml
```diff
@@ -1,5 +1,7 @@
 [mongodb]
 location = "mongodb://localhost:27017"
+username = "root"
+password = "root"
 
 [inx]
-address = "http://localhost:9029"
+address = "http://116.203.35.39:9029"
```
Alex6323, line 7:
```suggestion
address = "http://localhost:9029"
```

bin/inx-chronicle/src/broker.rs
```diff
@@ -1,52 +1,70 @@
...
-            }
-            InxEvent::LatestMilestone(milestone) => {
-                debug!("Received Milestone Event");
-                self.db.insert_milestone(milestone).await.unwrap()
-            }
-        }
-    }
+#[async_trait]
+impl Actor for Broker {
+    type Data = ();
```
Alex6323, line 36: So this `Data` is what the actor produces on initialization, right? Is there maybe a somewhat more descriptive name we could use instead? I find data to be a very vague term, and honestly, something I'ld expect to to constantly produced, not only on initialization, but maybe I am overlooking something. Since `handle_event` allows mutating this data, I wonder if `State` would be more appropriate. To me it makes sense, that the `init` fn would return the initial state of the actor. Wdyt?

bin/inx-chronicle/src/listener.rs
```diff
@@ -1,75 +1,198 @@
...
+    Runtime(#[from] RuntimeError),
 }
 
+#[derive(Debug)]
 pub struct InxListener {
-    // TODO: Consider storing `inx_client` to potentially restart some streams.
-    message_handle: JoinHandle<()>,
-    latest_milestone_handle: JoinHandle<()>,
+    config: InxConfig,
+    broker_handle: Addr<Broker>,
```
Alex6323, line 43: Since you named this field `broker_handle` and not `broker_addr` maybe consider renaming `Addr` to `Handle`.

bin/inx-chronicle/src/listener.rs
```diff
@@ -1,75 +1,198 @@
...
+}
+
+#[async_trait]
+impl Actor for InxListener {
+    type Data = InxClient<Channel>;
+    type Error = InxListenerError;
+
+    async fn init(&mut self, cx: &mut ActorContext<Self>) -> Result<Self::Data, Self::Error> {
+        info!("Connecting to INX...");
+        let mut inx_client = self.config.build().await?;
```
Alex6323, line 59: I still think something like `InxClient<Channel>::new(&self.config).await?` would be more future-proof.

bin/inx-chronicle/src/listener.rs
```diff
@@ -1,75 +1,198 @@
...
+
+        info!("Connected to INX.");
+        let response = inx_client.read_node_status(NoParams {}).await?;
+        info!("Node status: {:#?}", response.into_inner());
+
+        let message_stream = inx_client.listen_to_messages(MessageFilter {}).await?.into_inner();
+        let milestone_stream = inx_client.listen_to_latest_milestone(NoParams {}).await?.into_inner();
+
+        cx.spawn_actor_supervised::<InxStreamListener<inx::proto::Message>, _>(
```
Alex6323, line 68: Not necessary at all so ignore this if you don't like it: to reduce a bit of noise I'ld consider introducing those typedefs:
`type MessageStream = InxStreamListener<inx::proto::Message>;`
`type MilestoneStream = InxStreamListener<inx::proto::Milestone>;`

bin/inx-chronicle/src/listener.rs
```diff
@@ -1,75 +1,198 @@
...
+
+#[async_trait]
+impl HandleEvent<Report<InxStreamListener<inx::proto::Message>>> for InxListener {
+    async fn handle_event(
+        &mut self,
+        cx: &mut ActorContext<Self>,
+        event: Report<InxStreamListener<inx::proto::Message>>,
+        inx_client: &mut Self::Data,
+    ) -> Result<(), Self::Error> {
+        // TODO: Figure out why `cx.shutdown()` is not working.
```
Alex6323, line 88: Is that figured out now?

bin/inx-chronicle/src/listener.rs
```diff
@@ -1,75 +1,198 @@
...
+                }
+            },
         }
+        Ok(())
     }
+}
+
+pub struct InxStreamListener<I> {
+    broker_handle: Addr<Broker>,
+    _item: PhantomData<I>,
```
Alex6323, line 144: Probably too nit-picky and honestly I don't care, but maybe `_marker` or `_phantom` is a better field name? At least those are the names I see most often. This comment is more to make sure you chose this name deliberately. Close if you do! :slightly_smiling_face:

src/runtime/actor/context.rs
```diff
@@ -0,0 +1,109 @@
...
+    report::Report,
+    Actor,
+};
+use crate::runtime::{config::SpawnConfig, scope::RuntimeScope, shutdown::ShutdownStream};
+
+/// The context that an actor can use to interact with the runtime
+pub struct ActorContext<A: Actor> {
+    pub(crate) scope: RuntimeScope,
+    pub(crate) handle: Addr<A>,
+    pub(crate) receiver: ShutdownStream<Box<dyn Stream<Item = Envelope<A>> + Unpin + Send>>,
```
Alex6323, line 28: consider introducing the typedef
`type EnvelopeStream = Box<dyn Stream<Item = Envelope<A>> + Unpin + Send`;
because it reduces refactoring this definition to a one-line change.

src/runtime/actor/envelope.rs
```diff
@@ -0,0 +1,53 @@
...
+    ) -> Result<(), Self::Error>;
+}
+
+/// A dynamic event that can be sent to an actor which implements `HandleEvent` for it
+pub trait DynEvent<A: Actor>: Debug {
+    #[allow(missing_docs)]
+    fn handle<'a>(
+        self: Box<Self>,
+        cx: &'a mut ActorContext<A>,
+        act: &'a mut A,
```
Alex6323, line 28:
```suggestion
        actor: &'a mut A,
```
Maybe it's just me, but I don't like that `act` is an actual word itself.

src/runtime/actor/envelope.rs
```diff
@@ -0,0 +1,53 @@
...
+        data: &'a mut A::Data,
+    ) -> Pin<Box<dyn core::future::Future<Output = Result<(), A::Error>> + Send + 'a>>
+    where
+        Self: 'a,
+    {
+        act.handle_event(cx, *self, data)
+    }
+}
+
+/// Convenience type for boxed dynamic events
```
Alex6323, line 52:
```suggestion
/// Convenience type for boxed dynamic events.
```

src/runtime/actor/handle.rs
```diff
@@ -0,0 +1,81 @@
...
+use crate::runtime::{registry::ScopeId, scope::ScopeView};
+
+/// Error sending a message to an actor
+#[derive(Error, Debug)]
+#[error("Error sending message to actor: {0}")]
+pub struct SendError(String);
+
+#[allow(missing_docs)]
+impl SendError {
+    pub fn new<S: Into<String>>(msg: S) -> Self {
```
Alex6323, line 20:
```suggestion
    pub fn new(msg: impl Into<String>) -> Self {
```

src/runtime/actor/handle.rs
```diff
@@ -0,0 +1,81 @@
...
+
+impl<S: Into<String>> From<S> for SendError {
+    fn from(msg: S) -> Self {
+        Self::new(msg)
+    }
+}
+
+/// An actor handle, used to send events
+#[derive(Debug)]
+pub struct Addr<A: Actor> {
```
Alex6323, line 33: If this doc already uses the term `handle`, why not call this type `Handle` or even `ActorHandle`?

src/runtime/actor/handle.rs
```diff
@@ -0,0 +1,81 @@
...
+    }
+}
+
+impl<S: Into<String>> From<S> for SendError {
+    fn from(msg: S) -> Self {
+        Self::new(msg)
+    }
+}
+
+/// An actor handle, used to send events
```
Alex6323, line 31:
```suggestion
/// An actor handle, used to send events.
```

src/runtime/actor/handle.rs
```diff
@@ -0,0 +1,81 @@
...
+    pub(crate) scope: ScopeView,
+    pub(crate) sender: UnboundedSender<Envelope<A>>,
+}
+
+impl<A: Actor> Addr<A> {
+    pub(crate) fn new(scope: ScopeView, sender: UnboundedSender<Envelope<A>>) -> Self {
+        Self { scope, sender }
+    }
+
+    /// Shut down the actor with this handle. Use with care!
```
Alex6323, line 43: In this and the API docs below you use an imperative style except for the `is_closed` method. There is a guideline on this one can find online, and despite not being followed rigorously sometimes, especially when it comes to trait fns like for `Read` and `Write`, in Bee we decided to consistently use 3rd person singular in the docs according to that aforementioned API guideline. If you find those nit-picky things annoying then blame this on my german-ess please :trollface: Also please use proper punctuation. :walking:

src/runtime/actor/mod.rs
```diff
@@ -0,0 +1,53 @@
...
+#[async_trait]
+pub trait Actor: Send + Sync + Sized {
+    /// Custom data that is passed to all actor methods
+    type Data: Debug + Send + Sync;
+    /// Custom error type that is returned by all actor methods
+    type Error: Error + Send + Sync;
+
+    /// Set this actor's name, primarily for debugging purposes
+    fn name(&self) -> String {
+        std::any::type_name::<Self>().into()
```
Alex6323, line 32: why the allocation?

src/runtime/actor/report.rs
```diff
@@ -0,0 +1,39 @@
...
+pub type Report<A> = Result<SuccessReport<A>, ErrorReport<A>>;
+
+/// A report that an actor finished running with an error
+#[derive(Debug)]
+pub struct SuccessReport<A: Actor> {
+    /// The actor's state when it finished running
+    pub state: A,
+    /// The actor's data when it finished running
+    pub data: Option<A::Data>,
+}
```
Alex6323, line 16: okay, that's a conflict with what I suggested earlier about renaming `data` to `state`. You may not like this but I would rename `state` here to `actor` or something similar, because that's exactly what the type `A` says is stored in there. I know what you want to express here, and it also makes sense, so I guess it comes down to preferences again. An alternative could then be:
```rust
#[derive(Debug)]
pub struct SuccessReport<A: Actor> {
    pub stopped_actor: A,
    pub final_state: Option<A::State>,
}
```
I know it's a lot more wordy, but it constantly reminds the reader in whatever file s/he's in what's relevent to know about those fields without consulting the docs or jumping to this module.

Same suggestion goes for the ErrorReport struct.

src/runtime/config.rs
```diff
@@ -0,0 +1,56 @@
...
+
+use super::actor::{
+    envelope::{DynEvent, Envelope},
+    Actor,
+};
+
+/// Spawn configuration for an actor
+pub struct SpawnConfig<A> {
+    pub(crate) actor: A,
+    pub(crate) stream: Option<Box<dyn Stream<Item = Envelope<A>> + Unpin + Send>>,
```
Alex6323, line 14: use typedef `EnvelopeStream`

.gitignore
```diff
@@ -1,3 +1,5 @@
 /target
 
 **/.env
+
+/bin/inx-chronicle/config.toml
```
grtlr, line 5:
```suggestion
/bin/inx-chronicle/config.toml

```

bin/inx-chronicle/src/broker.rs
grtlr: If we don't have shared data, do we even need that type at all?

src/runtime/actor/mod.rs
```diff
@@ -0,0 +1,53 @@
...
+/// Module containing the actor address handle
+pub mod handle;
+/// Module containing actor exit report types
+pub mod report;
+
+/// The actor trait, which defines a task that is managed by the runtime
+#[async_trait]
+pub trait Actor: Send + Sync + Sized {
+    /// Custom data that is passed to all actor methods
+    type Data: Debug + Send + Sync;
```
grtlr, line 26: As an extension to my comment above: We probably still need this for everything that we initialize in `init`, or could this be removed?

src/runtime/actor/handle.rs
grtlr: I prefered the name `Addr`, this is probably the reason for the inconsistency. I think we should just rename this file to `addr.rs`, because I also always confuse it with the implementation of `HandleEvent`.

bin/inx-chronicle/config.example.toml
```diff
@@ -0,0 +1,7 @@
+[mongodb]
+location = "mongodb://localhost:27017"
+username = "root"
+password = "root"
+
+[inx]
+address = "http://116.203.35.39:9029"
```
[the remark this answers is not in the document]
Alexandcoats: oops

bin/inx-chronicle/src/broker.rs
Alexandcoats: Feels semantic at best. I consider the state to be the actor's struct. But the difference is more like externally initialized-state vs internally-initialized state.

bin/inx-chronicle/src/listener.rs
Alexandcoats: Hey it wasn't even my idea to call it `Addr`.

grtlr: `broker_addr` is fine :trollface:

bin/inx-chronicle/src/listener.rs
Alexandcoats: 👍

bin/inx-chronicle/src/listener.rs
Alexandcoats: no :(

It is a problem because the function borrows the `Self` struct and carries that borrow over an await point, but I need to be able to specify that it should only need to borrow one field of the struct.

bin/inx-chronicle/src/listener.rs
Alexandcoats: I chose it simply because it represents the stream's `Item`.

bin/inx-chronicle/src/listener.rs
```diff
@@ -1,75 +1,198 @@
...
+
+#[async_trait]
+impl HandleEvent<Report<InxStreamListener<inx::proto::Milestone>>> for InxListener {
+    async fn handle_event(
+        &mut self,
+        cx: &mut ActorContext<Self>,
+        event: Report<InxStreamListener<inx::proto::Milestone>>,
+        inx_client: &mut Self::Data,
+    ) -> Result<(), Self::Error> {
+        // TODO: Figure out why `cx.shutdown()` is not working.
```
grtlr, line 119: Same here.

src/runtime/actor/mod.rs
Alexandcoats: It is simply there to allow the actor a place to put data it creates without having a bunch of Options in it's own struct.

src/runtime/actor/report.rs
Alexandcoats: Made some naming changes I felt were appropriate. Let me know if you still hate it.

src/runtime/actor/mod.rs
Alexandcoats: Look what you have done. Now I've gone and had a `Cow`.

src/runtime/actor/handle.rs
Alexandcoats: no

bin/inx-chronicle/src/listener.rs
Alexandcoats: Take it up with @grtlr

bin/inx-chronicle/src/listener.rs
Alexandcoats: Not sure what this one is referencing 🤷

grtlr: The `// TODO` :trollface:

grtlr (approved):
LGTM 👍

grtlr (approved):

pvdrz (approved):

Alex6323 (approved):

PR #36, opened 2022-04-13
Rename `config.toml` to `config.example.toml`

grtlr:


Alex6323:
I am also doing that in the config PR I am preparing. Should we do the renaming as part of that PR instead? I also renamed it to `config.template.toml`, which I find a bit better that the example term as it's not really example "code".

grtlr:
Sure, sounds good!

PR #51, opened 2022-04-18
Add API

Alexandcoats:
This will likely be broken into a few PRs, but this is a proof of concept with lots of parts working together.


Cargo.toml
```diff
@@ -19,18 +19,86 @@ name = "inx-chronicle"
...
 log = { version = "0.4", default-features = false }
 mongodb = { version = "2.1", default-features = false, features = ["tokio-runtime"] }
 serde = { version = "1.0", features = ["derive"], default-features = false }
+serde_json = { version = "1.0", default-features = false, features = ["std"] }
 thiserror = { version = "1.0", default-features = false }
+time = { version = "0.3", default-features = false, features = ["std"] }
 tokio = { version = "1.17", default-features = false, features = [ "macros", "rt-multi-thread", "signal" ] }
 tokio-stream = { version = "0.1", default-features = false }
 toml = { version = "0.5", default-features = false }
+tonic = { version = "0.7", default-features = false }
```
grtlr, line 42: I'll add the missing types to the `inx` crate so that we don't need to import `tonic` directly.

bin/inx-chronicle/src/api/v1/extractors.rs
```diff
@@ -0,0 +1,77 @@
...
+use std::collections::HashMap;
+
+use async_trait::async_trait;
+use axum::extract::{FromRequest, Query};
+use hex::FromHex;
+use serde::{Deserialize, Serialize};
+
+use crate::api::error::APIError;
+
+#[derive(Clone, Default, Debug, Serialize, Deserialize)]
```
Alex6323, line 13: Just making sure: the default for `included` should be `false` for `MessagesQuery` but `true` for `OutputsQuery`?

Alexandcoats: Hmm. No, they should probably both be true.

bin/inx-chronicle/src/api/v1/routes.rs
```diff
@@ -0,0 +1,636 @@
...
+                .route("/included-message/:transaction_id", get(transaction_included_message)),
+        )
+        .route("/milestones/:index", get(milestone))
+        .nest("/analytics", Router::new().route("/addresses", get(address_analytics)))
+}
+
+async fn message(database: Extension<Database>, Path(message_id): Path<String>) -> APIResult<MessageResponse> {
+    let mut rec = database
+        .collection::<Document>("chrysalis_messages")
+        .find_one(doc! {"message_id": &message_id.to_string()}, None)
```
Alex6323, line 69: Isn't `MessageId` a `String` already?

bin/inx-chronicle/src/api/v1/routes.rs
```diff
@@ -0,0 +1,636 @@
...
+            .map(|m| m.as_string())
+            .collect::<Result<_, _>>()?,
+        payload: message.take_bson("payload").ok().map(Into::into),
+        nonce: message.convert_document::<U64, _>("nonce")?.into(),
+    })
+}
+
+async fn message_raw(database: Extension<Database>, Path(message_id): Path<String>) -> APIResult<Vec<u8>> {
+    let mut rec = database
+        .collection::<Document>("chrysalis_messages")
```
Alex6323, line 87: Should we put some `&str`s into constants? Especially those that are used several times?

Alexandcoats: yes

Cargo.toml
Alexandcoats: We need the tonic::transport::Error as well

Alexandcoats:
Split this into #53, #54, #55, #56.

PR #54, opened 2022-04-19
Add some bson helpers

Alexandcoats:


src/bson/ext.rs
```diff
@@ -0,0 +1,250 @@
...
+            _ => return Err(ValueAccessError::UnexpectedType),
+        })
+    }
+
+    fn as_u8(&self) -> Result<u8, ValueAccessError> {
+        Ok(match self {
+            Bson::Double(i) => *i as u8,
+            Bson::Boolean(i) => *i as u8,
+            Bson::Int32(i) => *i as u8,
+            Bson::Int64(i) => *i as u8,
```
Alex6323, line 57: I guess the circumstances in which those extension methods are used already ensure correct usage, i.e. that those won't panic? Or should we do manual bound checks and return an `ValueAccessError::OutOfBounds` or something to ensure the conversion is always possible? Just picked `as_u8` as an example here.

src/bson/ext.rs
```diff
@@ -0,0 +1,250 @@
...
+    Convert(#[from] mongodb::bson::de::Error),
+}
+
+#[allow(missing_docs)]
+pub trait DocPath {
+    fn split(self) -> Vec<String>;
+}
+
+impl DocPath for &str {
+    fn split(self) -> Vec<String> {
```
Alex6323, line 135: Maybe `parts` or `path_parts` could be a better name as `split` is already used for something more general.

src/bson/ext.rs
```diff
@@ -0,0 +1,250 @@
...
+    NullValue(String),
+    #[error(transparent)]
+    ValueAccess(#[from] ValueAccessError),
+    #[error(transparent)]
+    Convert(#[from] mongodb::bson::de::Error),
+}
+
+#[allow(missing_docs)]
+pub trait DocPath {
+    fn split(self) -> Vec<String>;
```
Alex6323, line 131: Oops. I should have placed the comment in line 135 here.

src/bson/ext.rs
```diff
@@ -0,0 +1,250 @@
...
+        }
+    }
+
+    fn convert_bson<T: DeserializeOwned, K: AsRef<str>>(&mut self, key: K) -> Result<T, DocError> {
+        Ok(from_bson(self.take_bson(key)?)?)
+    }
+
+    fn take_array(&mut self, key: impl AsRef<str>) -> Result<Vec<Bson>, DocError> {
+        Ok(self.take_bson(key)?.to_array()?)
+    }
```
Alex6323, line 190:
```suggestion
    }
    
```
:trollface:

pvdrz (changes requested):
you should seal both traits

src/bson/ext.rs
Alexandcoats: I think we absolutely could, but for the purpose of this helper trait it is unnecessary. The intention for this trait is to use it for values that are known.

src/bson/ext.rs
```diff
@@ -0,0 +1,259 @@
+// Copyright 2022 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+use mongodb::bson::{document::ValueAccessError, from_bson, from_document, Bson, Document};
+use sealed::sealed;
```
pvdrz, line 5: ohh nice, I always do it ol' school style

src/bson/mod.rs
```diff
@@ -36,11 +50,10 @@ impl From<u64> for U64 {
     }
 }
 
-#[allow(clippy::from_over_into)]
-impl Into<u64> for U64 {
-    fn into(self) -> u64 {
+impl From<U64> for u64 {
```
pvdrz, line 53: since when this is legal?

pvdrz (dismissed):

src/bson/mod.rs
grtlr: Wasn't it always?

pvdrz: yeah but we don't control the trait nor the type

src/bson/ext.rs
grtlr: Would it be annoying to do it the old school way? I don't like pulling in dependencies if it can be avoided.

grtlr (changes requested):

src/bson/ext.rs
Alexandcoats: 🤷 I think it's cleaner this way.

grtlr (approved):

pvdrz (approved):

PR #55, opened 2022-04-19
Add stardust api

Alexandcoats:
Dependent on #54

PR #61, opened 2022-04-20
Add api metrics

Alexandcoats:
Depends on #55


bin/inx-chronicle/src/api/metrics.rs
```diff
@@ -0,0 +1,163 @@
...
+    REGISTRY
+        .register(Box::new(RESPONSE_TIME_COLLECTOR.clone()))
+        .expect("Could not register collector");
+
+    REGISTRY
+        .register(Box::new(CONFIRMATION_TIME_COLLECTOR.clone()))
+        .expect("Could not register collector");
+}
+
+lazy_static! {
```
Alexandcoats, line 51: @grtlr 

> Would there be a way to implement this without using lazy_static? In bee-metrics we did something similar: https://github.com/iotaledger/bee/commit/8efaa02efe076da88a7d23ff1d1ace0aabe5c12a

Alex6323: The `once_cell` crate allows you to have lazy initialized global data without using a macro.

grtlr:
I think this work will be superseded by #123.

PR #88, opened 2022-04-25
Add SpawnActor event wrapper for spawning actors

Alexandcoats:
This wrapper simply spawns supervised actors, thus it can be used in combination with the `DelayEvent` type to spawn actors after a duration.

I also did a little cleanup of the error types.

### DIFF
--- a/src/runtime/actor/util.rs
+++ b/src/runtime/actor/util.rs
@@ -8,6 +8,7 @@ use async_trait::async_trait;
 use super::{
     context::ActorContext,
     event::{DynEvent, HandleEvent},
+    report::Report,
     Actor,
 };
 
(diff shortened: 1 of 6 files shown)


grtlr (dismissed):

src/runtime/actor/util.rs
```diff
@@ -47,3 +48,33 @@ where
         Ok(())
     }
 }
+
+/// An event which will spawn a supervised actor
```
Alex6323, line 52:
```suggestion
/// An event which will spawn a supervised actor.
```

src/runtime/actor/util.rs
```diff
@@ -47,3 +48,33 @@ where
...
+#[derive(Debug)]
+pub struct SpawnActor<A: Actor> {
```
Alex6323, line 54: Suggestion: `SpawnedActor` because when we handle that event it already happened, right?

Alexandcoats: No, the event will spawn the actor when it is processed.

Alex6323: Alright 👌 

And the event is fully pre-determined, meaning no other/user-defined code can be attached to it, such that it would be become more than just an actor spawner?

Alex6323 (approved):

PR #90, opened 2022-04-26
First version of `InxRequester`

grtlr:


Alex6323 (dismissed):

bin/inx-chronicle/src/inx/listener.rs
```diff
@@ -29,58 +25,50 @@ type MilestoneStream = InxStreamListener<inx::proto::Milestone>;
 
 #[derive(Debug, Error)]
 pub enum InxListenerError {
-    #[error(transparent)]
-    Inx(#[from] InxError),
+    // TODO Do we really need that error?
```
Alexandcoats, line 28: Only so we know the error is recoverable

bin/inx-chronicle/src/inx/worker.rs
```diff
@@ -0,0 +1,142 @@
...
+                    InxListenerError::SubscriptionFailed(_) => {
+                        cx.shutdown();
+                    }
+                    InxListenerError::Runtime(_) => {
+                        cx.shutdown();
+                    }
+                    InxListenerError::MissingBroker => {
+                        // If the handle is still closed, push this to the back of the event queue.
+                        // Hopefully when it is processed again the handle will have been recreated.
+                        if self.broker_addr.is_closed() {
```
Alexandcoats, line 79: There's a big problem with moving this logic here (which is solved by the address registry :trollface:). Since this actor doesn't supervise the broker, it will never overwrite the `broker_addr` it holds. So this event will be re-processed forever.

Alexandcoats:
LGTM other than the comments I made

Alex6323 (dismissed):

PR #110, opened 2022-04-29
Add Address Registry to the runtime

Alexandcoats:
This solves two problems:
1. The registry is now scoped along with the actors, so hierarchical design can be maintained.
2. Spawning actors now automatically adds the address to the registry (unless specified otherwise) so the `SpawnActor` event does not need to do anything special.

### DIFF
--- a/src/runtime/config.rs
+++ b/src/runtime/config.rs
@@ -1,6 +1,8 @@
 // Copyright 2022 IOTA Stiftung
 // SPDX-License-Identifier: Apache-2.0
 
+use std::fmt::Debug;
+
 use futures::{Stream, StreamExt};
 
 use super::actor::{
@@ -9,6 +11,7 @@ use super::actor::{
 };
 
 /// Spawn configuration for an actor.
+#[derive(Debug)]
 pub struct SpawnConfig<A> {
     pub(crate) actor: A,
     pub(crate) config: SpawnConfigInner<A>,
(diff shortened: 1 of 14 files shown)


src/runtime/config.rs
```diff
@@ -48,9 +99,14 @@ pub trait ConfigureActor: Actor {
...
     where
         S: 'static + Stream<Item = E> + Unpin + Send,
-        E: 'static + DynEvent<Self> + Send + Sync,
+        E: 'static + DynEvent<Self>,
     {
         SpawnConfig::<Self>::new(self).with_stream(stream)
     }
+
+    /// Sets whether the actor's address should be added to the registry.
+    fn add_to_registry(self, add: bool) -> SpawnConfig<Self> {
```
Alex6323, line 108: Idk but `with_registration(self, enable: bool)` might be nicer. Ignore if you disagree.

src/runtime/registry.rs
```diff
@@ -133,3 +160,28 @@ impl Deref for Scope {
         &self.inner
     }
 }
+
+#[derive(Debug, Default)]
+pub struct AddressRegistry {
```
Alex6323, line 165: Suggestion (ignore if you disagree): maybe `AddressRegistry` is too small of a type to name the module after it. My preference would be to either name this module `scope` and the other which is currently called "scope" `scope_view`, or even merge them into one `scope` module.

Alex6323 (approved):

src/runtime/registry.rs
Alexandcoats: It's not actually named after the address registry, but I agree it's kind of awkward. I want there to be separation between the public facing scope and the "registry" which is really just a linked tree of scopes. Maybe we could use `scope` instead of `registry` and `runtime_scope` instead of `scope`.

Alex6323: yea, that works for me :+1:

thibault-martinez (approved):

PR #111, opened 2022-04-29
Add Collector and Solidifier

Alexandcoats:
This is split off from the original PR, to limit the scope a little. It does depend on #110, however.


bin/inx-chronicle/src/collector/mod.rs
```diff
@@ -0,0 +1,275 @@
...
+}
+
+#[async_trait]
+impl Actor for Collector {
+    type State = HashMap<usize, Addr<Solidifier>>;
+    type Error = CollectorError;
+
+    async fn init(&mut self, cx: &mut ActorContext<Self>) -> Result<Self::State, Self::Error> {
+        let mut solidifiers = HashMap::new();
+        for i in 0..self.solidifier_count.max(1) {
```
Alex6323, line 52: I would take the `max` in the `new` ctor. Besides the minimum of `1` should we also define a maximum number of solidifiers as a constant?

Alex6323 (dismissed):

thibault-martinez (dismissed):

thibault-martinez (approved):

Alex6323 (approved):

PR #117, opened 2022-05-04
Add API config

Alexandcoats:


bin/inx-chronicle/src/api/config.rs
```diff
@@ -0,0 +1,45 @@
...
+        Self {
+            port: 9092,
+            allow_origins: Some(String::from("*").into()),
+        }
+    }
+}
+
+#[derive(Clone, PartialEq, Eq, Debug, Serialize, Deserialize, From)]
+#[serde(untagged)]
+pub enum SingleOrMultiple<T> {
```
pvdrz, line 29: could you add some docs for this type? I suspect that using `Vec<T>` instead and checking its length is not enough but I'm not sure why

pvdrz (changes requested):

Alex6323 (approved):

PR #118, opened 2022-05-04
[WIP] Add syncer

Alex6323:


Alexandcoats:
Couple of notes (I know this is still WIP tho)
- I would probably spawn the syncer at the top level from the launcher, as conceptually it has no connection to INX
- As I understand it, currently there is a single syncer which works its way from the earliest index forward forever with no stop condition. I would probably create a parent syncer which spawns lots of children to make requests in parallel chunks (just evenly split the total range of milestones). Their stop condition can be the completion of their chunk, and the parent's stop condition can be all of it's children exiting with `Ok`.
- If we want to rate-limit the requests, we should be able to simply sleep a little between each one.
- All the child syncers' work can probably be performed in `init`, since they won't necessarily need to process any events anyway. They can call `cx.shutdown()` at the beginning of `init` to ensure that they won't await the event loop.
- There is no need to hold both `start_index` in the struct and `current_index` in the internal state. Just rename `start_index` and mutate it.

PR #119, opened 2022-05-04
feat!(dto): introduce `dto`s to model data

grtlr:


Alex6323 (dismissed):

src/dto/mod.rs
```diff
@@ -0,0 +1,66 @@
+// Copyright 2022 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+#![allow(missing_docs)]
```
Alex6323, line 4: Is that temporary only (at the module level at least)? If so I would maybe add a TODO comment or something.

grtlr: Hmm, good question. Should we copy the documentation from `bee-message`? I'll create an issue for now.

Alex6323: I actually asked myself the same question. I think that would be the easiest.

Alex6323 (approved):

PR #120, opened 2022-05-05
Remove `Cow` in favor of `&'static str`

grtlr:


Alexandcoats:
I don't like this. It doesn't allow us to set it with `format!` if we want to. For instance, I want to override the Solidifier name fn so that it displays the solidifier ID.

PR #121, opened 2022-05-05
feat(syncer): add `Syncer` for requesting past milestones

Alex6323:
See the possible alternative #150 by @Alexandcoats 

From what I can tell and having reviewed it those are my favorite Pros, Neutrals and Cons of #150 :

Pros:
* is way more elegant and streamlined with the actor engine. I wonder why? xD
  (I will shamelessly port some stuff to my PR)
* also using a callback-channel makes the logic nicer

Neutral:
* PR 150's Syncer requires some setup phase (gathering the gaps in the database), this PR's Syncer doesn't need any setup: once it has determined the **correct** syncing range by using the node status and additional information it will start doing its thing; its unclear whether a setup phase might be problematic for big permanodes (it's eager vs lazy here)
* looping in event handlers with this actor framework is something one has to be careful about; PR 150 however uses only a very short for-loop over the gap ranges in the beginning and then progresses only in single steps, so that's no problem

Cons:

* uses the `latest_milestone_index` from the node status as the upper range target, but that's not enough (it needs to sync up
  to the first synced "live" milestone
* relies on channels, cloning actor address handles, and task spawning: the spawned tasks are not supervised, and together with the channel creation cause heap allocations; on the other hand ... this is a once-in-a-while effort so it might not even be a problem
* retries only the milestone request but it should also retry if the milestone couldn't be solidified; this PR here spends quite a few lines on this making the logic less pretty than PR 150; since we get a notification when a new SyncRecord was created and we know that syncing a milestone should never take longer than Coo's issuence rate, this Syncer knows when syncing a milestone really has failed, and needs to be retried. When it comes to such a situation this PR's Syncer would for simplicity sake check the whole range [first_failure_index..target_index]. This could be changed though at the cost of more complex logic EDIT: this is not the case anymore: I changed the logic so that only the failed milestone is retried.
* I don't see that resuming the syncer is as easy. It requires a new setup phase to build the new gaps list,
  and for this the logic has to be abstracted a bit (since instead of the node's pruning index it now should use
  the latest index it synced up to before and it needs to maintain that state somewhere), or correct me if I am wrong @Alexandcoats 

Additional:
* A performance bug in this PR (found by @Alexandcoats ) where the syncing would stall at some point has been fixed in the meantime.


bin/inx-chronicle/src/inx/worker.rs
```diff
@@ -48,19 +48,14 @@ impl Actor for InxWorker {
...
         log::info!("Connected to INX.");
-        let node_status = inx_client.read_node_status(NoParams {}).await?.into_inner();
 
-        if !node_status.is_healthy {
-            log::warn!("Node is unhealthy.");
-        }
-        log::info!("Node is at ledger index `{}`.", node_status.ledger_index);
-
-        cx.spawn_child(InxListener::new(inx_client.clone())).await;
+        // TODO: turn this back on!
```
Alexandcoats, line 57: do dis

Alex6323: hey we can always sync when it's part of the past :trollface:

bin/inx-chronicle/src/syncer.rs
```diff
@@ -0,0 +1,116 @@
...
+
+    async fn init(&mut self, cx: &mut ActorContext<Self>) -> Result<Self::State, Self::Error> {
+        // Send a `NodeStatus` request to the `InxWorker`
+        cx.addr::<InxWorker>().await.send(InxRequest::NodeStatus)?;
+        Ok(())
+    }
+}
+
+#[async_trait]
+impl HandleEvent<(MilestoneIndex, MilestoneIndex)> for Syncer {
```
Alexandcoats, line 91: Maybe make this event a struct for clarity.

Alexandcoats: Also there's really no reason to use the `MilestoneIndex` type if you do that.

bin/inx-chronicle/src/syncer.rs
```diff
@@ -0,0 +1,116 @@
...
+        let unsolid_milestones = self.collect_unsolid_milestones(*start_index, *end_index).await?;
+
+        let mut num_requested = 0;
+        for index in unsolid_milestones.into_iter() {
+            log::info!("Requesting milestone {}.", index);
+            cx.addr::<InxWorker>().await.send(InxRequest::milestone(index.into()))?;
+
+            num_requested += 1;
+
+            if num_requested == self.batch_size {
```
Alexandcoats, line 107: I still think that we should not batch these requests, but sleep between them instead.

Alexandcoats:
Rather that doing the `Syncer` -> `InxWorker` -> `Launcher` -> `Syncer` loop-de-loop, perhaps the `InxRequest` for `NodeStatus` should contain the sender's address?

Alex6323:
I don't know if it wasn't like that already but actually the data flow now is like:
`Launcher` -> `InxWorker` ->`Launcher` -> `Syncer`

bin/inx-chronicle/src/collector/mod.rs
```diff
@@ -155,7 +155,7 @@ pub mod stardust {
             message: inx::proto::Message,
             _solidifiers: &mut Self::State,
         ) -> Result<(), Self::Error> {
-            log::trace!("Received Stardust Message Event");
+            log::trace!("Received Stardust Message Event ({})", hex::encode(message.message_id.as_ref().unwrap().id.as_slice()));
```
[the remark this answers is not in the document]
Alex6323: Done

bin/inx-chronicle/src/collector/mod.rs
```diff
@@ -176,7 +176,10 @@ pub mod stardust {
             metadata: inx::proto::MessageMetadata,
             _solidifiers: &mut Self::State,
         ) -> Result<(), Self::Error> {
-            log::trace!("Received Stardust Message Referenced Event");
+            log::trace!(
```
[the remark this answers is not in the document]
Alex6323: Done

bin/inx-chronicle/src/collector/mod.rs
```diff
@@ -200,7 +203,10 @@ pub mod stardust {
             milestone: inx::proto::Milestone,
             solidifiers: &mut Self::State,
         ) -> Result<(), Self::Error> {
-            log::trace!("Received Stardust Milestone Event");
+            log::trace!(
```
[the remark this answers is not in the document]
Alex6323: Done

Alexandcoats:
> uses the latest_milestone_index from the node status as the upper range target, but that's not enough (it needs to sync up
to the first synced "live" milestone
The intention with spawning it under the listener is that there should be no logic necessary to check whether we have started solidifying, as the existence of the milestone stream should be enough to prove that we have. However, that may not actually work as well as I had hoped.

> retries only the milestone request but it should also retry if the milestone couldn't be solidified
I don't think we should. If it couldn't solidify the first time, why would it the second time? If there's a network issue the whole inx listener will restart so we'll retry it anyway since this is a child of that.

> It requires a new setup phase to build the new gaps list

I'll assume that by resuming you mean requesting the next milestone after one completes solidification, in which case it is no different from yours. We only need to re-request gaps if the whole actor restarts, which should only happen on things like network errors, in which case it is best to check the gaps again anyway.

PR #123, opened 2022-05-05
feat: add initial support for metrics

pvdrz:


bin/inx-chronicle/src/metrics.rs
```diff
@@ -0,0 +1,59 @@
...
+        let addr = "0.0.0.0:6969".parse().unwrap();
+        let fut = tokio::spawn(serve_metrics(addr, registry.clone()).map(|res| res.unwrap()));
+
+        Ok((registry, fut))
+    }
+
+    async fn shutdown(&mut self, cx: &mut ActorContext<Self>, (_, fut): &mut Self::State) -> Result<(), Self::Error> {
+        log::debug!("{} shutting down ({})", self.name(), cx.id());
+
+        // FIXME: is this good enough to stop the task?
```
Alexandcoats, line 34: Does bee_metrics support a graceful shutdown for this service?

Alexandcoats:
Can we add some basic metrics to this PR?

bin/inx-chronicle/src/metrics.rs
pvdrz: nope

bin/inx-chronicle/src/main.rs
```diff
@@ -12,6 +12,7 @@ mod collector;
 mod config;
 #[cfg(feature = "inx")]
 mod inx;
+mod metrics;
```
Alex6323, line 15: I guess the plan is to also feature-gate this?

bin/inx-chronicle/src/metrics.rs
Alexandcoats: Big oof. Is there anyone we can ask about whether there is a better way?

bin/inx-chronicle/src/main.rs
Alexandcoats: Should be

bin/inx-chronicle/src/metrics.rs
pvdrz: Is this that bad? The call to abort is non-blocking so the worst it could happen is that we will keep serving metrics until the process ends.

Cargo.toml
```diff
@@ -24,6 +24,7 @@ path = "bin/inx-chronicle/src/main.rs"
 # Required
 async-recursion = { version = "1.0", default-features = false }
 async-trait = { version = "0.1", default-features = false }
+bee-metrics = { git = "https://github.com/iotaledger/bee", branch = "mainnet-develop-0.4", default-features = false, optional = true }
```
Alexandcoats, line 27: Move this to the a new section?

Cargo.toml
```diff
@@ -90,3 +91,6 @@ stardust = [
     "dep:bee-message-stardust",
     "dep:bee-rest-api-stardust",
 ]
+metrics = [
```
Alexandcoats, line 94: Alphabetical order

bin/inx-chronicle/src/metrics.rs
Alexandcoats: No, it's probably fine.

bin/inx-chronicle/src/metrics.rs
```diff
@@ -0,0 +1,59 @@
...
+#[async_trait]
+impl Actor for MetricsWorker {
+    type State = (Arc<Registry>, JoinHandle<()>);
+
+    type Error = Infallible;
+
+    async fn init(&mut self, _cx: &mut ActorContext<Self>) -> Result<Self::State, Self::Error> {
+        let registry = Arc::new(Registry::default());
+
+        // FIXME: pass address via config.
```
Alexandcoats, line 24: Might as well add this to the PR

pvdrz:
> Can we add some basic metrics to this PR?

I added the CPU and memory usage metrics.

bin/inx-chronicle/src/metrics.rs
```diff
@@ -0,0 +1,59 @@
...
+        let addr = "0.0.0.0:6969".parse().unwrap();
+        let fut = tokio::spawn(serve_metrics(addr, registry.clone()).map(|res| res.unwrap()));
```
Alexandcoats: We should await this future, so I would spawn a task to do that and send a shutdown signal. See the API actor for an example.

Alexandcoats:
> > Can we add some basic metrics to this PR?
> 
> I added the CPU and memory usage metrics.

Can we add channel size as well? Perhaps a guage which increments when `send` is called and decrements when a message is received?

pvdrz:
> > > Can we add some basic metrics to this PR?
> > 
> > 
> > I added the CPU and memory usage metrics.
> 
> Can we add channel size as well? Perhaps a guage which increments when `send` is called and decrements when a message is received?

we don't have that merged in `bee_metrics` so that will have to wait.

bin/inx-chronicle/src/main.rs
```diff
@@ -218,6 +225,22 @@ impl HandleEvent<Report<ApiWorker>> for Launcher {
...
+#[cfg(feature = "metrics")]
+#[async_trait]
+impl HandleEvent<Report<metrics::MetricsWorker>> for Launcher {
+    async fn handle_event(
+        &mut self,
+        cx: &mut ActorContext<Self>,
+        _event: Report<metrics::MetricsWorker>,
+        _config: &mut Self::State,
+    ) -> Result<(), Self::Error> {
+        // FIXME: what to do here?
```
Alexandcoats, line 237: I would probably only shutdown on panic. Otherwise I would restart the actor.

Alex6323 (dismissed):

Alexandcoats (approved):

grtlr (approved):

PR #128, opened 2022-05-06
chore: Fix `inx` and `api` features

Alexandcoats:
Fixed the CI defs so that the "common" feature groups are more accurate. Did some blanket testing of feature combos as well.


bin/inx-chronicle/src/stardust_inx/error.rs
```diff
@@ -3,7 +3,7 @@
 
 use thiserror::Error;
 
-use super::listener::InxListenerError;
+use super::listener::StardustInxListenerError;
```
grtlr, line 6: I wonder if that should just be `stardust::InxListenerError`?

bin/inx-chronicle/src/config.rs
```diff
@@ -9,8 +9,8 @@ use thiserror::Error;
 
 #[cfg(feature = "api")]
 use crate::api::ApiConfig;
-#[cfg(feature = "inx")]
-use crate::inx::InxConfig;
+#[cfg(all(feature = "stardust", feature = "inx"))]
+use crate::stardust_inx::StardustInxConfig;
```
grtlr, line 13: Same here (maybe `stardust::InxConfig`)?

bin/inx-chronicle/src/stardust_inx/error.rs
Alexandcoats: I renamed them so we wouldn't have potential naming conflicts later in the event that we do add more protocols that use INX.

grtlr:
> I renamed them so we wouldn't have potential naming conflicts later in the event that we do add more protocols that use INX. 

Yeah, that's also fine. I guess if we ever want to support other protocols, we could think about defining a trait that abstracts over that anyways.

grtlr (dismissed):

grtlr:
Superseded by #128.

grtlr (approved):

PR #143, opened 2022-05-11
fix: `unreachable_pub` instances and add compiler warning

Alex6323:
I like keeping things as private as possible.


src/dto/output/alias.rs
```diff
@@ -4,7 +4,7 @@
 use bee_message_stardust::output as stardust;
 use serde::{Deserialize, Serialize};
 
-pub use super::{
+pub(crate) use super::{
```
grtlr, line 7: I don't know why this was `pub` in the first place.

```suggestion
use super::{
```

Alexandcoats (dismissed):

Alexandcoats (approved):

PR #149, opened 2022-05-12
chore(launcher): move launcher into own module

Alex6323:


Alexandcoats (dismissed):

bin/inx-chronicle/src/api/error.rs
```diff
@@ -41,19 +41,25 @@ pub enum ApiError {
     BadParse(ParseError),
     #[error("Invalid time range")]
     BadTimeRange,
+    // TODO: use or remove dead code
```
grtlr, line 44: Do you plan on doing these `TODO`s in this PR?

bin/inx-chronicle/src/launcher.rs
```diff
@@ -0,0 +1,243 @@
+// Copyright 2022 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+//! Contains the `Launcher` actor.
```
grtlr, line 4: While we're at it, maybe we can write a sentence or two about the purpose of this actor.

bin/inx-chronicle/src/main.rs
```diff
@@ -10,248 +10,17 @@ mod cli;
...
-                ActorError::Panic | ActorError::Aborted => {
-                    cx.shutdown();
-                }
-            },
-        }
-
-        Ok(())
-    }
-}
+use launcher::Launcher;
```
grtlr, line 23: Prefix these two `use` statements with `self::` for future-proof formatting.

bin/inx-chronicle/src/api/error.rs
Alex6323: I'm not sure because I didnt introduce those variants myself. Would have to talk to @AlexandCoats about it first, and see if they are needed at some point.

Alexandcoats: These were part of the indexation API, so they can be removed.

Alexandcoats (dismissed):

PR #160, opened 2022-05-16
chore(ci): re-enable `deny` via `RUSTFLAGS`

grtlr:


grtlr:
This can't be merged until the dependencies, namely `petgraph` and `indexmap` are updated.

Alex6323 (dismissed):

workingjubilee:
This is not an issue with the indexmap or petgraph crates per se,
so you may be waiting a long time if you expect them to fix this.
I recommend familiarizing yourself with these discussions:
https://github.com/bluss/indexmap/issues?q=is%3Aissue+arguments

grtlr:
@workingjubilee Thank you very much for your help! This is a very confusing situation.

workingjubilee:
This may require a `[patch.crates-io]` override instead.

grtlr:
Thank you so much, will try it out!

grtlr:
Ah, also does not work:

```
warning: patch for `indexmap` uses the features mechanism. default-features and features will not take effect because the patch dependency does not support this mechanism
```

workingjubilee:
Oh right. Bother. Tedious...

grtlr:
All good, thank you so much anyway for taking the time respond and think about this!

grtlr:
I'm closing this for now, maybe these things will settle in the future.

Issue #167, opened 2022-05-17
Enabled metrics should be configurable via features


grtlr:
There are two different kinds of metrics, the ones we use for debugging our implementation and ones that are used in production to monitor the node.

Alexandcoats:
Would it perhaps be better to have multiple features for metrics, since adding them usually requires different data structures (see the channels) and this may be best determined at compile time.

grtlr:
I agree, similar to `tokio-console`, the channel-based metrics should not be exposed to the user. So they can be put behind a feature flag.

In general, I think we should be careful adding new features because they can become unwieldy.

PR #170, opened 2022-05-17
feat(metrics): add solidification counter metric

Alexandcoats:


grtlr (approved):

grtlr:
Before we merge these metrics, we should do #167 to make it easier to track which metrics are enabled under which circumstances. @pvdrz

Alex6323 (approved):

grtlr (approved):

PR #163, opened 2022-05-16
fix!: rename `Block` and update `inx`

grtlr:
Todos:

* [x] Update dependencies
  * [x] https://github.com/iotaledger/bee/pull/1407
  * [x] https://github.com/iotaledger/inx/pull/42
* [x] Update Hornet configuration

PR #172, opened 2022-05-18
fix(db): Rename `message_id` to `_id` 

grtlr:


pvdrz (dismissed):

Alex6323 (dismissed):

bin/inx-chronicle/src/api/responses.rs
```diff
@@ -34,8 +34,16 @@ pub struct InfoResponse {
 
 impl_success_response!(InfoResponse);
 
-#[derive(Clone, Debug, Serialize, Deserialize)]
-pub struct SyncDataResponse(pub SyncData);
+/// An aggregation type that represents the ranges of completed milestones and gaps.
```
Alexandcoats, line 37: This is needed for the syncer, so relegating it to the API is no bueno

grtlr: I think this is resolved for now :trollface:

Alexandcoats: No, we still need it in the syncer

grtlr:
I went ahead and merged this, we can add move it later again if we need to, but my goal is to merge #163 too.

PR #176, opened 2022-05-19
fix: update `bee-metrics` and log first error for process metrics

pvdrz:


bin/inx-chronicle/src/metrics.rs
```diff
@@ -89,9 +89,13 @@ impl Actor for MetricsWorker {
 
         let process_metrics_handle = {
             tokio::spawn(async move {
+                if let Err(e) = metrics.update().await {
```
Alexandcoats, line 92: Shouldn't the task just exit in this case?

pvdrz: Not sure how likely is to have a one time error that never repeats like `proc` files being unavailable. To be honest I never expected this to fail at all.

Alexandcoats: I'm confused. If this fails the first time, won't it just fail over and over in the loop?

pvdrz: I am not sure. I'd believe that it is possible that `update` could fail the first time and not fail later but I don't have any scenario to prove it either.

Alexandcoats (dismissed):

bin/inx-chronicle/src/metrics.rs
```diff
@@ -89,10 +89,24 @@ impl Actor for MetricsWorker {
...
+                            retries = MAX_RETRIES;
+                        }
+                        Err(e) => {
+                            log::warn!("Cannot update process metrics: {e}");
+                            retries -= 1;
+                        }
+                    }
                 }
+
+                log::warn!("Could not update process metrics after {MAX_RETRIES} retries");
```
grtlr, line 109:
```suggestion
                log::warn!("Could not update process metrics after {MAX_RETRIES} retries.");
```

grtlr (dismissed):

Alexandcoats (dismissed):

grtlr (approved):

Alex6323 (approved):

PR #186, opened 2022-05-24
fix(db)!: store `Address` instead of `AliasAddress` in `Unlock`

grtlr:


src/db/model/stardust/block/output/unlock_condition.rs
```diff
@@ -137,11 +140,18 @@ pub(crate) mod test {
...
+        let block = get_test_immut_alias_address_condition(get_test_alias_address_as_address());
         let bson = to_bson(&block).unwrap();
         assert_eq!(block, from_bson::<UnlockCondition>(bson).unwrap());
     }
 
+    pub(crate) fn get_test_alias_address_as_address() -> Address {
+        bee_block_stardust::address::Address::from(bee_block_stardust::address::AliasAddress::new(
+            bee_test::rand::output::rand_alias_id(),
+        ))
+        .into()
```
Alex6323, line 152: Would be a bit more readable if those types were imported (maybe even just within this function with `bee_block_stardust::address::*`).

Alex6323 (approved):

Alexandcoats (approved):

PR #194, opened 2022-05-24
fix(docker): fix `Dockerfile`

grtlr:


bin/inx-chronicle/config.template.toml
```diff
@@ -4,7 +4,7 @@ username = "root"
 password = "root"
 
 [api]
-port = 9092
+port = 8042
```
Alexandcoats, line 7: Arbitrary change?

grtlr: It was colliding with the Rest API of Hornet's `private_tangle`.

PR #195, opened 2022-05-24
feat: bring back `Scope::spawn_task`

pvdrz:


src/runtime/task/mod.rs
```diff
@@ -0,0 +1,21 @@
...
+pub(super) mod error;
+pub(super) mod report;
+
+use std::{borrow::Cow, error::Error};
+
+use async_trait::async_trait;
+
+#[async_trait]
+pub trait Task: Send + Sync + Sized {
+    type Error: Error + Send + Sync;
```
Alexandcoats, line 13: Shouldn't need to be Sync

src/runtime/scope.rs
```diff
@@ -212,6 +214,53 @@ impl RuntimeScope {
...
+    pub async fn spawn_task<T, Sup>(&mut self, mut task: T, supervisor_addr: Addr<Sup>) -> AbortHandle
+    where
+        T: Task + 'static,
+        Sup: 'static + HandleEvent<TaskReport<T>>,
+    {
+        let (abort_handle, abort_registration) = AbortHandle::new_pair();
+        let mut child_scope = self.child(None, Some(abort_handle.clone())).await;
+        let child_task = spawn_task(task.name().as_ref(), async move {
+            let fut = task.run();
+            let res = Abortable::new(AssertUnwindSafe(fut).catch_unwind(), abort_registration).await;
```
Alexandcoats, line 227: Can you combine these two lines?

bin/inx-chronicle/src/metrics.rs
```diff
@@ -28,7 +29,7 @@ impl MetricsWorker {
 
 pub struct MetricsState {
     server_handle: (JoinHandle<()>, Option<oneshot::Sender<()>>),
-    process_metrics_handle: JoinHandle<()>,
+    abort_handles: Vec<AbortHandle>,
```
Alexandcoats, line 32: Don't actually need to store these if you don't want to. You can just use `cx.abort().await` in the parent actor's shutdown fn.

pvdrz: So I don't actually need to return them in the `spawn_task` method?

Alexandcoats: It is not strictly necessary, but I would do it anyway because it could be useful.

src/runtime/scope.rs
```diff
@@ -212,6 +218,50 @@ impl RuntimeScope {
...
+            child_scope.abort().await;
```
Alexandcoats, line 232: We probably shouldn't abort the children on exit, similar to how I remove this from the actor, as this allows the task to spawn children if it wants to and manage their lifecycles better.

Alexandcoats (approved):

Alexandcoats:
🤷

PR #204, opened 2022-05-25
feat: add manual actor name impls

Alexandcoats:


bin/inx-chronicle/src/api/mod.rs
```diff
@@ -131,6 +131,10 @@ impl Actor for ApiWorker {
         log::info!("Stopping API server");
         run_result
     }
+
+    fn name(&self) -> std::borrow::Cow<'static, str> {
```
Alex6323, line 135: question: the name stays the same whole lifetime of the app ('static), right, so why make this a Cow rather than just return a `&'static str`?

Alex6323: Forget it....I see that it sometimes needs to own a dynamically created value.

Alex6323 (dismissed):

pvdrz (approved):

grtlr (approved):

PR #230, opened 2022-05-31
feat(types): add Copy and Into<Bson> impls

Alexandcoats:


src/types/ledger/inclusion_state.rs
```diff
@@ -18,6 +21,22 @@ pub enum LedgerInclusionState {
...
+        }
+    }
+}
+
+impl From<LedgerInclusionState> for Bson {
+    fn from(val: LedgerInclusionState) -> Self {
+        val.to_string().into()
+    }
+}
+
```
grtlr, line 39: The `Display` implementation is error prone, because it could get out of sync with the `#[serde(rename = "...")]`. Maybe it's better to do the following:

```suggestion
impl From<LedgerInclusionState> for Bson {
    fn from(val: LedgerInclusionState) -> Self {
        // Panic: This can never fail because we implement `Deserialize` on the type.
        mongodb::bson::to_bson(&val).unwrap()
    }
}

```

src/types/stardust/block/address/mod.rs
```diff
@@ -50,6 +51,17 @@ impl FromStr for Address {
...
+    fn from(val: Address) -> Self {
+        match val {
+            Address::Ed25519(a) => doc! {"ed25519": &a},
+            Address::Alias(a) => doc! {"alias": &a},
+            Address::Nft(a) => doc! {"nft": &a},
+        }
+        .into()
+    }
+}
+
```
grtlr, line 64: Same problems as with `LedgerInclusionState`.

src/types/stardust/block/output/mod.rs
```diff
@@ -64,6 +65,16 @@ impl FromStr for OutputId {
...
+impl From<OutputId> for Bson {
+    fn from(val: OutputId) -> Self {
+        doc! {
+            "transaction_id": &val.transaction_id,
+            "index": &(val.index as i32),
+        }
+        .into()
+    }
+}
+
```
grtlr, line 77: And again

grtlr (approved):

Alex6323 (approved):

PR #238, opened 2022-05-31
Add global lock for tests using `mongodb`

pvdrz:


Alexandcoats:
Since this is for testing, it seems like it would be better to just use `lazy_static`, as it will be simpler.

pvdrz:
> Since this is for testing, it seems like it would be better to just use `lazy_static`, as it will be simpler.

why simpler?

pvdrz:
> Sure, but this is for testing only, and lazy_static is simpler.
> 
> That being said, I think the approach isn't what we want here anyway. I don't think we should be locking tests, but rather creating a wrapper for the database that will drop the database on exit so that tests clean up after themselves.

what if two tests use the db at the same time?

Alexandcoats:
It should not matter if two tests use it at the same time, but if it ever does then we would want a lock for that.

> if one test drops the db and cleans it while the other is using it?

This would be an Rc or Arc so that it would only drop when there are no remaining refs.

Alexandcoats:
Superceded by #567

PR #243, opened 2022-06-01
feat(inx): retry on INX connection errors

Alex6323:
Adds back knocking on Hornet's door every 5s for all eternity.

### DIFF
--- a/bin/inx-chronicle/src/launcher.rs
+++ b/bin/inx-chronicle/src/launcher.rs
@@ -89,7 +89,11 @@ impl HandleEvent<Report<super::stardust_inx::InxWorker>> for Launcher {
             }
             Report::Error(report) => match report.error {
                 ActorError::Result(e) => match e {
-                    InxError::ConnectionError(_) | InxError::TransportFailed(_) => {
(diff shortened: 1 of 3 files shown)


Alexandcoats:
This should delay a `SpawnActor` event, not a custom one.

bin/inx-chronicle/src/launcher.rs
```diff
@@ -75,15 +80,20 @@ impl HandleEvent<Report<super::stardust_inx::InxWorker>> for Launcher {
...
     ) -> Result<(), Self::Error> {
+        use super::stardust_inx::InxError;
         match event {
             Report::Success(_) => {
                 cx.abort().await;
             }
             Report::Error(report) => match report.error {
                 ActorError::Result(e) => match e {
-                    super::stardust_inx::InxError::MongoDb(e) => match e.kind.as_ref() {
+                    InxError::ConnectionError(_) | InxError::TransportFailed(_) => {
```
grtlr, line 92: We should only retry on `InxError::ConnectionError`, if `TransportFailed` we should report an `ERROR` and shutdown Chronicle.

PR #244, opened 2022-06-01
feat!: database improvements and cleanup

Alexandcoats:


src/db/collections/block.rs
```diff
@@ -177,85 +199,123 @@ impl MongoDb {
...
+            .aggregate(
+                vec![
+                    doc! { "$match": { "block.payload.transaction_id": &output_id.transaction_id } },
+                    doc! { "$replaceRoot": { "newRoot": {
+                        "output": { "$arrayElemAt": [ "$block.payload.essence.outputs", &(output_id.index as i64) ] } ,
+                        "metadata": {
+                            "output_id": &output_id,
+                            "block_id": "$block_id",
+                            "transaction_id": "$block.payload.transaction_id",
+                            "booked": "$metadata.milestone_index",
```
Alex6323, line 237: can the "spent" field can be ignored?

src/db/collections/block.rs
```diff
@@ -177,85 +199,123 @@ impl MongoDb {
...
+            .transpose()?;
+
+        Ok(output)
+    }
+
+    /// Get an [`OutputWithMetadata`] by [`OutputId`].
+    pub async fn get_output_with_metadata(&self, output_id: &OutputId) -> Result<Option<OutputWithMetadata>, Error> {
+        let mut output: Option<OutputWithMetadata> = self
+            .0
+            .collection::<Output>(BlockDocument::COLLECTION)
```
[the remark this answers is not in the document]
Alexandcoats: It doesn't actually matter due to the aggregation pipeline, but you're right that it would be clearer.

src/db/collections/block.rs
```diff
@@ -1,65 +1,84 @@
...
+                    )
+                    .build(),
+                None,
+            )
+            .await?;
+
+        collection
+            .create_index(
+                IndexModel::builder()
+                    .keys(doc! { "block.payload.transaction_id": 1 })
```
Alex6323, line 66: is it not "block.payload.transaction.id" ?
(`Block` > `Payload` > `TransactionPayload` > `TransactionId`)

src/db/collections/block.rs
Alexandcoats: It does.

src/db/collections/block.rs
```diff
@@ -276,15 +345,15 @@ impl MongoDb {
...
+        start_milestone: crate::types::tangle::MilestoneIndex,
+        end_milestone: crate::types::tangle::MilestoneIndex,
     ) -> Result<Option<AddressAnalyticsResult>, Error> {
         Ok(self.0.collection::<BlockDocument>(BlockDocument::COLLECTION)
         .aggregate(
             vec![
                 doc! { "$match": {
                     "inclusion_state": LedgerInclusionState::Included,
-                    "milestone_index": { "$gt": start_milestone, "$lt": end_milestone },
+                    "metadata.referenced_by_milestone_index": { "$gt": start_milestone, "$lt": end_milestone },
```
Alex6323, line 356: that is a non-inclusive range, right? Are we sure about that?

Alexandcoats: Changed this while adding the paging state

bin/inx-chronicle/src/api/error.rs
```diff
@@ -81,12 +81,17 @@ impl IntoResponse for ApiError {
 
 #[derive(Error, Debug)]
 pub enum ParseError {
-    #[error(transparent)]
-    Bool(#[from] ParseBoolError),
+    #[allow(dead_code)]
```
Alex6323, line 84: remove?

bin/inx-chronicle/src/api/error.rs
```diff
@@ -100,7 +105,7 @@ pub struct ErrorBody {
     #[serde(skip_serializing)]
     status: StatusCode,
     code: u16,
-    block: String,
+    message: String,
```
Alex6323, line 108: that's funny actually

src/db/collections/ledger_update.rs
```diff
@@ -19,19 +20,21 @@ use crate::{
 };
 
 /// Contains all informations related to an output.
+#[allow(missing_docs)]
```
Alex6323, line 23: remove now?

src/db/collections/ledger_update.rs
```diff
@@ -94,25 +108,116 @@ impl MongoDb {
...
+                .0
+                .collection::<LedgerUpdateDocument>(LedgerUpdateDocument::COLLECTION)
+                .aggregate(
+                    vec![
+                        doc! { "$match": { "at.milestone_index": { "$gt": start_milestone, "$lt": end_milestone } } },
+                        doc! { "$facet": {
+                            "total": [
+                                { "$group" : {
+                                    "_id": "$address",
+                                    "transfers": { "$count": { } }
```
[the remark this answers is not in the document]
Alex6323: same in below cases

bin/inx-chronicle/src/api/error.rs
Alexandcoats: Yeah oops lol

src/db/collections/ledger_update.rs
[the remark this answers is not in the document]
Alexandcoats: https://www.mongodb.com/docs/manual/reference/operator/aggregation/count-accumulator/#mongodb-group-grp.-count

 they are the same 🤷

src/db/collections/ledger_update.rs
Alexandcoats: No, because then I would have to document the fields and I don't want to.

Alex6323: right

PR #264, opened 2022-06-13
feat(docker)!: save MongoDB in `volume`

### LINKED ISSUE #227: Mount `volume` for database in `docker-compose.hornet.yml`
Right now the database is only stored within the `mongo` container, that might not be persisted. It should probably look something like this:

```yml
mongo:
  image: mongo
  restart: always
  command: --quiet --logpath /dev/null
  volume:
    - ./data:/data/db
```

That would probably also require to set the permissions of the folder on the host.

grtlr:
Closes #227.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,11 +11,16 @@ The easiest way to start Chronicle is by using our supplied Dockerfile.
 First you need to setup the correct permissions for the node's database:
 
 ```sh
-mkdir docker/alphanet
-
+mkdir docker/hornet_data
 groupadd -g 65532 nonroot
 useradd -g nonroot -u 65532 nonroot
-chown nonroot:nonroot docker/alphanet/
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -11,11 +11,16 @@ The easiest way to start Chronicle is by using our supplied Dockerfile.
 First you need to setup the correct permissions for the node's database:
 
 ```sh
-mkdir docker/alphanet
-
+mkdir docker/hornet_data
```
Alex6323, line 14: should we really bind ourselves to `hornet`? What about `node_data`?

grtlr: This is strongly tied to `docker/docker-compose.hornet.yml`, we will have to change documentation for Bee anyways.

Alex6323 (approved):

PR #271, opened 2022-06-14
fix(runtime): use `warn!` instead of `error!`

### LINKED ISSUE #269: Fix INX `failed to establish connection` error

grtlr:
Closes #269.


Alexandcoats:
Why?

grtlr:
The supervisor of actors should be responsible for omitting errors, depending on the context its context. The runtime, as a library, should not report anything to the user.

Alexandcoats (approved):

PR #281, opened 2022-06-14
feat(api): add JWT authentication

Alexandcoats:


bin/inx-chronicle/config.template.toml
```diff
@@ -7,6 +7,9 @@ database_name = "chronicle"
 [api]
 port = 8042
 allow_origins = "0.0.0.0"
+# password = "password"
+password_hash = "c42cf2be3a442a29d8cd827a27099b0c"
```
grtlr, line 11:
```suggestion
password_hash = "c42cf2be3a442a29d8cd827a27099b0c" # "password"
```

bin/inx-chronicle/config.template.toml
```diff
@@ -7,6 +7,9 @@ database_name = "chronicle"
...
+password_salt = "saltines"
```
grtlr, line 12: Shouldn't that be a hash too?

bin/inx-chronicle/src/api/auth.rs
```diff
@@ -0,0 +1,80 @@
...
+use super::{config::ApiData, ApiError};
+
+pub struct Auth;
+
+#[async_trait]
+impl<B: Send> FromRequest<B> for Auth {
+    type Rejection = ApiError;
+
+    async fn from_request(req: &mut axum::extract::RequestParts<B>) -> Result<Self, Self::Rejection> {
+        // Unwrap: Infallable
```
grtlr, line 22: Could you please use `// Panic: ...` and explain why this is infallible?

bin/inx-chronicle/src/api/config.rs
```diff
@@ -13,17 +17,62 @@ use super::error::ConfigError;
...
 impl Default for ApiConfig {
     fn default() -> Self {
         Self {
             port: 9092,
-            allow_origins: Some(String::from("*").into()),
+            allow_origins: Some("*".to_string().into()),
+            password_hash: "c42cf2be3a442a29d8cd827a27099b0c".to_string(),
+            password_salt: "saltines".to_string(),
+            jwt_salt: "Chronicle".to_string(),
+            jwt_expiration: time::Duration::minutes(30).try_into().unwrap(),
```
grtlr, line 36: Is that really a sensible default? Shouldn't the expiration be optional too?

bin/inx-chronicle/src/api/error.rs
```diff
@@ -12,17 +15,23 @@ use thiserror::Error;
 #[derive(Error, Debug)]
 #[allow(missing_docs)]
 pub enum InternalApiError {
+    #[cfg(feature = "stardust")]
+    #[error(transparent)]
+    BeeStardust(#[from] bee_block_stardust::Error),
     #[error(transparent)]
```
grtlr, line 21: ❤️ the alphabetical sorting

bin/inx-chronicle/src/launcher.rs
```diff
@@ -32,6 +32,8 @@ pub struct LauncherState {
     config: ChronicleConfig,
     #[allow(dead_code)]
     db: MongoDb,
+    #[cfg(feature = "api")]
+    secret_key: libp2p_core::identity::ed25519::SecretKey,
```
grtlr, line 36: Do we really need an `ed25519` secret key, or could we go with something more common? Maybe Bee just uses `ed25519` because the node requires it anyways?

bin/inx-chronicle/config.template.toml
Alexandcoats: No? We don't check the salt against an unhashed salt.

bin/inx-chronicle/src/api/auth.rs
Alexandcoats: "Infallable" IS the explanation, because `<OriginalUri as FromRequest>::Error = Infallable`

But I will add that to the comment.

bin/inx-chronicle/src/api/config.rs
Alexandcoats: If they never expire, then anyone who gets their hands on a token could use it forever. Not sure how important that is, but I'd rather default to expiring and let users decide if they want that risk.

bin/inx-chronicle/src/launcher.rs
Alexandcoats: We can use pretty much anything. What do you suggest?

grtlr: Ok, I did some more digging and `ed25519` seems to be a good choice, sorry about that. We should, however, use a more basic library, such as `ed25519-dalek` instead of `libp2p_core` (which has bitten us from time to time).

bin/inx-chronicle/config.template.toml
```diff
@@ -7,6 +7,8 @@ database_name = "chronicle"
 [api]
 port = 8042
 allow_origins = "0.0.0.0"
+password_hash = "c42cf2be3a442a29d8cd827a27099b0c" # "password"
+password_salt = "saltines"
```
grtlr, line 11:
```suggestion
password_salt = "Chronicle"
```

bin/inx-chronicle/config.template.toml
```diff
@@ -7,6 +7,8 @@ database_name = "chronicle"
 [api]
 port = 8042
 allow_origins = "0.0.0.0"
+password_hash = "c42cf2be3a442a29d8cd827a27099b0c" # "password"
```
grtlr, line 10: We should also explicitly add an expiration field for the JWT here. By default the expiration should be set to never expire, with a comment to explain that this should be changed in production.

bin/inx-chronicle/src/api/config.rs
```diff
@@ -13,17 +17,63 @@ use super::error::ConfigError;
...
     fn default() -> Self {
         Self {
             port: 9092,
-            allow_origins: Some(String::from("*").into()),
+            allow_origins: Some("*".to_string().into()),
+            password_hash: "c42cf2be3a442a29d8cd827a27099b0c".to_string(),
+            password_salt: "saltines".to_string(),
+            jwt_salt: "Chronicle".to_string(),
+            // 30 min
+            jwt_expiration: Duration::from_secs(30 * 60),
```
grtlr, line 37: With the changes to the `config.template.toml`, we can make the expiration optional and set it to `None` by default.

bin/inx-chronicle/src/api/config.rs
grtlr: If we change the `config.template.toml` to include the expiration field (set to never expire) and write a comment we should be good. Securing Chronicle in production is not our job but rather has to be done individually by users — we just need to give them the tools they need to do so, together with documentation.

bin/inx-chronicle/src/api/config.rs
```diff
@@ -13,17 +17,63 @@ use super::error::ConfigError;
...
 
 impl Default for ApiConfig {
     fn default() -> Self {
         Self {
             port: 9092,
-            allow_origins: Some(String::from("*").into()),
+            allow_origins: Some("*".to_string().into()),
+            password_hash: "c42cf2be3a442a29d8cd827a27099b0c".to_string(),
+            password_salt: "saltines".to_string(),
+            jwt_salt: "Chronicle".to_string(),
```
grtlr, line 35: What's the difference between the `password_salt` and the `jwt_salt`?

bin/inx-chronicle/src/api/mod.rs
```diff
@@ -41,18 +46,18 @@ pub type ApiResult<T> = Result<T, ApiError>;
 #[derive(Debug)]
 pub struct ApiWorker {
     db: MongoDb,
-    config: ApiConfig,
+    config: ApiData,
```
Alex6323, line 49: Should we rename the field to `api_data` in order to avoid confusion with `ApiConfig` or maybe rename `ApiData` to `ApiConfigData`?

bin/inx-chronicle/src/api/routes.rs
```diff
@@ -15,7 +22,37 @@ pub fn routes() -> Router {
...
-    Router::new().nest("/api", router).fallback(not_found.into_service())
+    Router::new()
+        .route("/login", post(login))
+        .nest("/api", router.route_layer(from_extractor::<Auth>()))
+        .fallback(not_found.into_service())
+}
+
+#[derive(Deserialize)]
+struct LoginInfo {
+    #[serde(rename = "password")]
```
Alex6323, line 33: Can be removed?

bin/inx-chronicle/src/config.rs
```diff
@@ -9,8 +9,17 @@ use thiserror::Error;
...
+    #[error(transparent)]
+    Api(#[from] crate::api::ConfigError),
     #[error("failed to read file: {0}")]
     FileRead(std::io::Error),
+    #[cfg(feature = "api")]
+    #[error("failed to decode key: {0}")]
+    KeyDecode(libp2p_core::identity::error::DecodingError),
+    #[cfg(feature = "api")]
+    #[error("failed to read key bytes")]
+    KeyRead,
```
Alex6323, line 22: Do those need to be separate from the `Api` error? (They have the same feature gate, so I wonder if those variants would either be better placed as variants of `crate::api::ConfigError`, or if we should remove the api feature gate?

bin/inx-chronicle/config.template.toml
Alexandcoats: aw, but I like saltines

bin/inx-chronicle/config.template.toml
Alexandcoats: Still not sure about the rationale of defaulting to never expire, but I'll add the field.

bin/inx-chronicle/src/api/config.rs
Alex6323: There's no reason to let the token never expire (if we set the default to 24 hours it should be enough in almost all cases). Also ... it's just about not needing to re-authenticate oneself for subsequent requests. So why open a security hole for a convenience thing. Chronicle will just create a new token for them. And also... most people don't read comments carefully enough or understand some of the implications if they don't change the 'never expire' default.

Alex6323 (approved):

Alex6323 (approved):

bin/inx-chronicle/src/api/mod.rs
```diff
@@ -142,3 +147,47 @@ async fn shutdown_signal(recv: oneshot::Receiver<()>) {
...
+    /// If the bytes do not constitute a valid Ed25519 secret key, an error is
+    /// returned.
+    pub fn from_bytes(mut sk_bytes: impl AsMut<[u8]>) -> Result<SecretKey, ed25519_dalek::SignatureError> {
+        use zeroize::Zeroize;
+        let sk_bytes = sk_bytes.as_mut();
+        let secret = ed25519_dalek::SecretKey::from_bytes(&*sk_bytes)?;
+        sk_bytes.zeroize();
+        Ok(SecretKey(secret))
+    }
+}
```
grtlr, line 193: This should probably go into a separate file.

bin/inx-chronicle/src/launcher.rs
```diff
@@ -58,28 +64,56 @@ impl Actor for Launcher {
...
+            let secret_key = match &cli_args.identity {
+                Some(path) => keypair_from_file(path)?,
+                None => {
+                    if let Ok(path) = std::env::var("IDENTITY_PATH") {
+                        keypair_from_file(&path)?
+                    } else {
+                        crate::api::SecretKey::generate()
+                    }
+                }
+            };
```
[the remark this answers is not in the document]
Alex6323: Introduce?
https://github.com/iotaledger/inx-chronicle/blob/main/bin/inx-chronicle/src/config.rs#L24

bin/inx-chronicle/src/api/config.rs
```diff
@@ -1,29 +1,96 @@
...
+            public_routes: Default::default(),
+            identity_path: None,
         }
     }
 }
 
+#[derive(Clone, Debug)]
+pub struct ApiData {
+    pub port: u16,
+    pub allow_origins: Option<SingleOrMultiple<String>>,
```
grtlr, line 45: Do we still need the `Option<>` here? I can't really think of a use-case right now.

grtlr (dismissed):

PR #284, opened 2022-06-14
feat(runtime)!: allow adding streams to own event loop

Alexandcoats:
This PR changes the way adding additional streamed events to actor event loops works. Previously, the supervisor was required to construct the stream and add it via `SpawnConfig`. This put too much of the child actor's logic in the hands of the supervisor. This PR adds the `add_stream` method to the `ActorContext`, which allows actors to add streams they construct to their own event loop.


src/runtime/merge.rs
```diff
@@ -6,96 +6,77 @@ use std::{
...
+        let me = self.project();
 
-        if a_first {
-            poll_next(me.a, me.b, cx)
-        } else {
-            poll_next(me.b, me.a, cx)
+        let n = me.streams.len();
+        let idx = *me.idx;
+        *me.idx += 1;
+        *me.idx %= n;
```
Alex6323, line 57: I think I'ld prefer:
`*me.idx = (*me.idx + 1) % n`

src/runtime/merge.rs
```diff
@@ -6,96 +6,77 @@ use std::{
...
-        Merge::new(self, other)
+        Merge::new(self).merge(other)
     }
 }
 impl<S: ?Sized> MergeExt for S where S: Stream {}
 
 /// Stream returned by the [`merge`](MergeExt::merge) method.
 #[pin_project::pin_project]
-pub(crate) struct Merge<T, U> {
+pub(crate) struct Merge<I> {
```
Alex6323, line 25: You are generalizing this whole Stream `Merge` logic to merge multiple streams, right? This feels a bit like a separate feature PR.

Alex6323 (approved):

Alex6323 (approved):

PR #302, opened 2022-06-20
fix(api)!: fix endpoint prefixes

grtlr:


bin/inx-chronicle/src/api/error.rs
```diff
@@ -42,6 +42,8 @@ pub enum ApiError {
     NoResults,
     #[error("No endpoint found")]
     NotFound,
+    #[error("Endpoint not implented")]
```
Alexandcoats, line 45: Implented

Alex6323 (approved):

PR #314, opened 2022-06-21
fix(api)!: TIP compliance for `history` API fields

### LINKED ISSUE #312: Make `LedgerUpdateResponse` fully TIP-compliant
* [x] `paging_state` should be `cursor`
* [x] `transaction_id` and `output_index` should be `outputId`
* [x] `data` should be `items`
* [x] All fields should use camelCase

grtlr:
Closes #312.

### DIFF
--- a/bin/inx-chronicle/src/api/stardust/history/responses.rs
+++ b/bin/inx-chronicle/src/api/stardust/history/responses.rs
@@ -10,18 +10,16 @@ use crate::api::impl_success_response;
 #[derive(Clone, Debug, Serialize, Deserialize)]
 pub struct TransactionHistoryResponse {
     pub address: String,
-    pub transactions: Vec<Transfer>,
-    pub paging_state: Option<String>,
+    pub items: Vec<Transfer>,
+    pub cursor: Option<String>,
 }
 
 impl_success_response!(TransactionHistoryResponse);
 
 #[derive(Clone, Debug, Serialize, Deserialize)]
 pub struct Transfer {
-    #[serde(rename = "transactionId")]
-    pub transaction_id: String,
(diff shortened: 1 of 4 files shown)


Alexandcoats (approved):

bin/inx-chronicle/src/api/stardust/history/responses.rs
```diff
@@ -10,18 +10,16 @@ use crate::api::impl_success_response;
 #[derive(Clone, Debug, Serialize, Deserialize)]
 pub struct TransactionHistoryResponse {
     pub address: String,
-    pub transactions: Vec<Transfer>,
-    pub paging_state: Option<String>,
+    pub items: Vec<Transfer>,
```
Alex6323, line 13: Since the elements of the vec are `Transfer`s, why not call this field `transfers` then? :trollface:

bin/inx-chronicle/src/api/stardust/history/routes.rs
```diff
@@ -68,8 +67,8 @@ async fn transaction_history(
         .collect::<Result<_, ApiError>>()?;
 
     Ok(TransactionHistoryResponse {
-        transactions,
+        items: transactions,
```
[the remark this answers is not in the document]
grtlr: But `transactions` is already ambiguous :trollface:, because the list contains outputs. I think we should do a renaming path once the TIP has settled.

bin/inx-chronicle/src/api/stardust/history/responses.rs
grtlr: Because @Alexandcoats came up with `Transfer`, but we could not agree on it yet. 😁

PR #336, opened 2022-06-22
fix(api): add ledger index to output queries

### LINKED ISSUE #319: Return proper `ledger_index` in `OutputMetadataResponse`
The field `OutputMetadataResponse::ledger_index` describes:

> The current ledger index for which the request was made.

To properly set this field we need to store the latest `MilestoneIndex` from `proto::ListenToLedgerUpdates` in a Chronicle-wide state object that is also shared with the API and retrieved from there. Alternatively, we could also perform a more complex DB query.

Alexandcoats:
Fixes #319


src/db/collections/block.rs
```diff
@@ -313,6 +320,15 @@ impl MongoDb {
...
                     } },
+                    doc! { "$lookup": {
+                        "from": MilestoneDocument::COLLECTION,
+                        "pipeline": [
+                            { "$sort": { "milestone_timestamp": -1 } },
+                            { "$limit": 1 }
+                        ],
+                        "as": "latest_milestone"
+                    } },
+                    doc! { "$set": { "latest_milestone": { "$first": "$latest_milestone" } } },
```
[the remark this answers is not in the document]
Alexandcoats: Not a bad idea, but out of the scope of this PR. Want to create an issue?

src/db/collections/block.rs
```diff
@@ -313,6 +320,15 @@ impl MongoDb {
                         "block.payload.transaction_id": &output_id.transaction_id,
                         "$expr": { "$gt": [{ "$size": "$block.payload.essence.outputs" }, &(output_id.index as i64)] }
                     } },
+                    doc! { "$lookup": {
+                        "from": MilestoneDocument::COLLECTION,
+                        "pipeline": [
+                            { "$sort": { "milestone_timestamp": -1 } },
+                            { "$limit": 1 }
```
Alex6323, line 327: out of interest: why not just use the `milestone_index` (protocol internal clock)? Due to the indexing there's no performance difference, right, and it becomes a matter of taste?

Alex6323 (dismissed):

src/types/ledger/output_metadata.rs
```diff
@@ -31,6 +31,7 @@ pub struct OutputMetadata {
     pub transaction_id: TransactionId,
     pub booked: MilestoneIndexTimestamp,
     pub spent: Option<SpentMetadata>,
+    pub latest_milestone: Option<MilestoneIndexTimestamp>,
```
grtlr, line 34: Why is this field optional? There should always be a `latest_milestone` by design, right?

src/db/collections/block.rs
grtlr: We use the `milestone_index` in most places, so it might make sense to unify. Ultimately, it's up to you @Alexandcoats.

src/types/ledger/output_metadata.rs
```diff
@@ -39,40 +40,41 @@ pub struct OutputWithMetadata {
...
 
 #[cfg(feature = "inx")]
-impl From<inx::LedgerSpent> for OutputWithMetadata {
+impl From<inx::LedgerSpent> for LedgerUpdate {
     fn from(value: inx::LedgerSpent) -> Self {
-        let mut output_with_metadata = OutputWithMetadata::from(value.output);
+        let mut delta = LedgerUpdate::from(value.output);
 
-        output_with_metadata.metadata.spent = Some(SpentMetadata {
-            transaction_id: value.transaction_id_spent.into(),
```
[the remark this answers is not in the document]
Alexandcoats: We don't actually store that anywhere in the ledger updates table, so it's not needed here...unless you think we should store that.

src/db/collections/milestone.rs
```diff
@@ -269,8 +269,35 @@ impl MongoDb {
...
+        &self,
+        index: MilestoneIndex,
+    ) -> Result<Option<MilestoneIndexTimestamp>, Error> {
+        let mut stream = self
+            .0
+            .collection::<MilestoneIndexTimestamp>(MilestoneDocument::COLLECTION)
+            .find(
+                doc! { "milestone_index": { "$gte": index } },
+                FindOptions::builder().sort(doc! {"milestone_index": 1}).build(),
+            )
```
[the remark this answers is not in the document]
grtlr: I'm not sure if this is the right collection to get the information from. I think it would be better to have a `ledger_index` field in the `NodeStatus` collection, which we update separately. This way we could also achieve _O(1)_ lookups (instead of _O(n)_ for the `$max`-based approach).

When we start using `ReadUnspentOutputs` we can't really add this information to the `MilestoneDocument` anyway.

Cargo.lock
```diff
@@ -141,9 +141,9 @@ dependencies = [
 
 [[package]]
 name = "axum"
-version = "0.5.11"
+version = "0.5.9"
```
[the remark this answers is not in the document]
Alexandcoats: Accidental

src/db/collections/outputs.rs
```diff
@@ -0,0 +1,246 @@
...
+            let output = self
+                .0
+                .collection::<OutputWithMetadataResult>(OutputDocument::COLLECTION)
+                .aggregate(
+                    vec![
+                        doc! { "$match": {
+                            "output_id": &output_id,
+                            "metadata.booked.milestone_index": { "$lte": ledger_index }
+                        } },
+                        doc! { "$set": {
```
Alexandcoats, line 146: Change to return spent metadata even if it is beyond our ledger index

```
if spent != null
  ledger_index = max(ledger_index, spent.milestone_index)
```

src/db/collections/outputs.rs
```diff
@@ -0,0 +1,246 @@
...
+            let metadata = self
+                .0
+                .collection::<OutputMetadataResult>(OutputDocument::COLLECTION)
+                .aggregate(
+                    vec![
+                        doc! { "$match": {
+                            "output_id": &output_id,
+                            "metadata.booked.milestone_index": { "$lte": ledger_index }
+                        } },
+                        doc! { "$set": {
```
Alexandcoats, line 186: Same as above

bin/inx-chronicle/src/stardust_inx/mod.rs
```diff
@@ -122,6 +123,17 @@ impl Actor for InxWorker {
...
+        log::info!("Reading unspent outputs.");
+        let mut unspent_output_stream = inx_client.read_unspent_outputs(NoParams {}).await?.into_inner();
+
+        let mut updates = Vec::new();
+        while let Some(unspent_output) = unspent_output_stream.try_next().await? {
+            let ledger_output: inx::LedgerOutput = unspent_output.output.unwrap().try_into()?;
+            updates.push(ledger_output.into());
+        }
+        log::info!("Inserting {} unspent outputs.", updates.len());
+        self.db.insert_ledger_updates(updates).await?;
```
Alex6323, line 135: I think I remember from the call that first buffering all ledger updates in a Vec happens on purpose, but I am not sure and I am not sure whether it's obvious. Could you add a comment why we're not inserting them immediatedly?

src/db/collections/status.rs
```diff
@@ -24,14 +25,43 @@ impl MongoDb {
...
+    pub async fn get_ledger_index(&self) -> Result<Option<MilestoneIndex>, Error> {
+        self.0
+            .collection::<StatusDocument>(StatusDocument::COLLECTION)
+            .find_one(doc! {}, None)
+            .await
+            .map(|doc| doc.and_then(|doc| doc.ledger_index))
+    }
+
+    /// Sets the current ledger index.
+    pub async fn set_ledger_index(&self, ledger_index: MilestoneIndex) -> Result<(), Error> {
```
Alex6323, line 55: This is - admittedly - a bit nit-picky, but `set_` to me implies it will actually be set to the given value which is not the case. A better name might be `update_ledger_index`. The doc for this method should probably also explain that the ledger index gets updated only if it marks a new max.

src/db/collections/status.rs
```diff
@@ -24,14 +25,43 @@ impl MongoDb {
...
     pub async fn set_network_name(&self, network_name: String) -> Result<(), Error> {
         self.0
             .collection::<StatusDocument>(StatusDocument::COLLECTION)
-            .insert_one(StatusDocument { network_name }, None)
+            .update_one(
+                doc! {},
+                doc! { "$set": { "network_name": network_name } },
+                UpdateOptions::builder().upsert(true).build(),
+            )
+            .await?;
```
Alex6323, line 40: Isn't that a problem that this `StatusDocument` can be changed non-atomically? What if I run `Chronicle` in a different network, but the ledger index is still the old? So what I am asking is ... should setting the network name not reset the ledger index?

bin/inx-chronicle/src/stardust_inx/mod.rs
Alexandcoats: Because they come from a stream, not an iterator.

src/db/collections/status.rs
Alexandcoats: Fair

src/db/collections/status.rs
Alex6323: right, should this method fail then?

src/db/collections/ledger_update.rs
```diff
@@ -108,52 +108,30 @@ impl MongoDb {
...
+                    output_id: delta.metadata.output_id,
+                    at: delta.metadata.spent.map(|s| s.spent).unwrap_or(delta.metadata.booked),
+                    is_spent: delta.metadata.spent.is_some(),
+                })?;
+                self.0
                     .collection::<LedgerUpdateDocument>(LedgerUpdateDocument::COLLECTION)
                     .update_one(
-                        doc! { "output_id": metadata.output_id, "is_spent": is_spent },
-                        doc! { "$setOnInsert": bson::to_document(&ledger_update_document)? },
+                        doc.clone(),
```
Alex6323, line 133: is MongoDB smart enough to use the index here? Even if so ... why clone the whole struct and not use the contents of the index? (the struct might get bigger in the future)

Alexandcoats:
Created #412 to track a potential change proposed by @Alex6323

PR #339, opened 2022-06-22
feat(api): implement `is_healthy` check for `health/` API endpoint

### LINKED ISSUE #307: Implement `is_healthy` check for `health/` API endpoint
Currently, the `is_healthy` check in the `api` module always reports `true`. However, if we are connected to INX, we should also check (via INX) if the node is healthy too. This would require adding an INX connection to the `ApiWorker`.

We should also think about other meaningful health criteria for Chronicle.

DaughterOfMars:
Fixes #307

### DIFF
--- a/bin/inx-chronicle/src/stardust_inx/mod.rs
+++ b/bin/inx-chronicle/src/stardust_inx/mod.rs
@@ -206,8 +206,11 @@ impl HandleEvent<IsHealthy> for InxWorker {
         healthy &= node_status.is_healthy;
 
         let latest_inserted_ms = self.db.get_latest_milestone().await?;
-        healthy &= latest_inserted_ms.map_or(false, |latest_inserted_ms| {
-            node_status.confirmed_milestone.milestone_info.milestone_index == latest_inserted_ms
+        healthy &= latest_inserted_ms.map_or(false, |MilestoneIndex(latest_inserted_ms)| {
+            // If the latest confirmed ms from the node is either the last ms we inserted or the next one
+            // (because we are still working on it) then we are healthy
+            (latest_inserted_ms..=latest_inserted_ms + 1)
+                .contains(&node_status.confirmed_milestone.milestone_info.milestone_index)
         });
 
         let first_ms = node_status.tangle_pruning_index + 1;


bin/inx-chronicle/src/api/mod.rs
```diff
@@ -68,6 +68,7 @@ impl Actor for ApiWorker {
         let api_handle = cx.handle().clone();
         let port = self.config.port;
         let routes = routes()
+            .layer(Extension((&***cx).clone()))
```
Alex6323, line 71: A comment would be nice with what type we end up with via this deref chain (ScopeView?)

bin/inx-chronicle/src/util.rs
```diff
@@ -0,0 +1,35 @@
+// Copyright 2022 IOTA Stiftung
```
Alex6323, line 1: Consider naming this module `health.rs` or `health_check.rs` which is more specific.

bin/inx-chronicle/src/stardust_inx/mod.rs
```diff
@@ -192,3 +193,18 @@ impl HandleEvent<Report<Syncer>> for InxWorker {
...
+impl HandleEvent<IsHealthy> for InxWorker {
+    async fn handle_event(
+        &mut self,
+        _cx: &mut ActorContext<Self>,
+        IsHealthy(sender): IsHealthy,
+        inx_client: &mut Self::State,
+    ) -> Result<(), Self::Error> {
+        let node_status = NodeStatus::try_from(inx_client.read_node_status(NoParams {}).await?.into_inner())
+            .map_err(InxError::InxTypeConversion)?;
+        sender.send(node_status.is_healthy).ok();
```
Alex6323, line 207: Additionally we should check the `latest_confirmed_milestone` or whatever it was exactly called (from the node status) against our database (that also ensures that we - at the moment of the request - are still connected to our mongodb instance and that it's responsive).

grtlr: Sounds like a good idea!

src/runtime/actor/addr.rs
```diff
@@ -95,3 +94,15 @@ impl<A: Actor> Deref for OptionalAddr<A> {
         &self.0
     }
 }
+
+impl<A: Actor> Clone for OptionalAddr<A> {
+    fn clone(&self) -> Self {
+        Self(self.0.clone())
+    }
+}
```
grtlr, line 102: Why can't this be derived?

Alexandcoats: The derive imposes `A: Clone` bounds.

grtlr: Ah gottcha

bin/inx-chronicle/src/stardust_inx/mod.rs
```diff
@@ -192,3 +193,34 @@ impl HandleEvent<Report<Syncer>> for InxWorker {
...
+        let node_status = NodeStatus::try_from(inx_client.read_node_status(NoParams {}).await?.into_inner())
+            .map_err(InxError::InxTypeConversion)?;
+        let mut healthy = true;
+        healthy &= node_status.is_healthy;
+        let latest_inserted_ms = self.db.get_latest_milestone().await?;
+        healthy &= if let Some(latest_inserted_ms) = latest_inserted_ms {
+            node_status.confirmed_milestone.milestone_info.milestone_index == latest_inserted_ms
+        } else {
+            false
+        };
```
Alex6323, line 214: I don't care so much, but maybe using `map_or` here is a bit more concise.

bin/inx-chronicle/src/check_health.rs
```diff
@@ -0,0 +1,35 @@
...
+
+#[async_trait]
+impl CheckHealth for ScopeView {
+    async fn is_healthy<A: Actor>(&self) -> Result<bool, CheckHealthError>
+    where
+        A: 'static + HandleEvent<IsHealthy>,
+    {
+        let (sender, receiver) = tokio::sync::oneshot::channel();
+        self.addr::<A>().await.send(IsHealthy(sender))?;
+        Ok(receiver.await?)
```
Alex6323, line 33: You something possible hang due to Chronicle being unhealthy making this await point block forever? Or in other words: Should we race this await point against a health-check timeout which yields false if the health check wasn't performed in time?

bin/inx-chronicle/src/api/routes.rs
```diff
@@ -19,22 +19,32 @@ pub fn routes() -> Router {
...
         name: "Chronicle".into(),
         version,
         is_healthy,
     }
 }
 
-pub async fn health() -> StatusCode {
-    match is_healthy() {
+pub async fn health(Extension(scope): Extension<ScopeView>) -> StatusCode {
+    match is_healthy(&scope).await {
```
Alex6323, line 47: I'ld prefer classic `if ... else` on a boolean tbh, but I leave it up to you.

bin/inx-chronicle/src/check_health.rs
Alexandcoats: I've thought about it, because currently hornet does block forever due to the deadlock issue, but tbh that's not something we should have to account for. Hornet needs to fix their mess.

bin/inx-chronicle/src/stardust_inx/mod.rs
```diff
@@ -192,3 +193,36 @@ impl HandleEvent<Report<Syncer>> for InxWorker {
...
+        let node_status = NodeStatus::try_from(inx_client.read_node_status(NoParams {}).await?.into_inner())
+            .map_err(InxError::InxTypeConversion)?;
+
+        let mut healthy = true;
+        healthy &= node_status.is_healthy;
+
+        let latest_inserted_ms = self.db.get_latest_milestone().await?;
+        healthy &= latest_inserted_ms.map_or(false, |latest_inserted_ms| {
+            node_status.confirmed_milestone.milestone_info.milestone_index == latest_inserted_ms
+        });
```
grtlr, line 214: Shouldn't this be a range, rather than a concrete comparison? What would happen if the node is currently still processing the cone of the milestone in `node_status`?

bin/inx-chronicle/src/stardust_inx/mod.rs
```diff
@@ -192,3 +193,36 @@ impl HandleEvent<Report<Syncer>> for InxWorker {
...
+
+        let first_ms = node_status.tangle_pruning_index + 1;
```
grtlr, line 216: I'm not sure at the moment, is the pruning index still included by the node, or the latest index that has been pruned away?

Alex6323: the latter (the above calculated index points to the first prunable milestone)

bin/inx-chronicle/src/stardust_inx/mod.rs
grtlr: We should check for `last_inserted_ms >= confirmed_milestone_index - 3` (or some other magical number). Just remember that the milestone index can also be 0 or 1.

Alexandcoats: What's the actual issue with returning unhealthy once in a blue moon due to this?

grtlr: If it's easily avoidable, we should do it. We don't know how people will use this endpoint, or what kind of infra they build on top of it.

Probably it's fine to check against the current and the previous milestone though.

Alexandcoats: Makes sense to me

grtlr (approved):

PR #359, opened 2022-06-29
fix(db): insert `Block`s using transaction

### LINKED ISSUE #348: Use transactions and batch insert to write cones to the database
@Alexandcoats wrote:

> The validity of this index is defined by its existence in the database, meaning that the milestone was successfully added. However, because we are not using transactions, there are several things that can go wrong.
>
> 1. We request an output which was inserted, but the milestone was not yet. Very slim chance, but technically possible. In this case the query would likely fail with a 500 error because we expect the value to exist.
> 2. We request an output which is unspent on milestone N, spent on N+X. The data returned may reflect the spent status, though the milestone was not yet inserted. Again, this would probably return a 500 error. Alternatively, we may get the unspent data but N+X ledger index because the data and milestone could be inserted between the first match and the lookup. This would not return an error in the API, but would instead return invalid results.
>
> Because there is no alignment between reads and writes, we cannot trust some of our queries to function properly. For instance, the query I wrote for getting the ledger index returns invalid data half of the time because it contains lookups and data can be inserted between them.

We should use MongoDB's transactions together with a batch insert to write milestone cone streams to the database.

Please note that this does not solve all problems with the ledger index, as outlined in #347.

grtlr:
Closes #348.


grtlr:
We probably won't be needing transaction because of #336, so I'm closing this for now.

PR #366, opened 2022-06-30
fix(platform): support shutdown in Docker environment

### LINKED ISSUE #365: Graceful shutdown of docker container not possible
When trying to stop Chronicle gracefully via `docker-compose down -t100` the contained gets killed after the 100 seconds.

We devops would like to avoid this because it's not a clean shutdown and we don't know if shutting down with timeout could damage (/ lead to inconsistencies) the database.

According to the documentation:
```
Compose stop attempts to stop a container by sending a SIGTERM. It then waits for a [default timeout of 10 seconds](https://docs.docker.com/engine/reference/commandline/compose_stop/). After the timeout, a SIGKILL is sent to the container to forcefully kill it. If you are waiting for this timeout, it means that your containers aren’t shutting down when they receive the SIGTERM signal.
```

So I think, Chronicle would need a SIGTERM handler to shut itself down in a clean way.

Alex6323:
This is shamelessly copied from Bee. Honors go to Adam who implemented it there. Closes #365.
(not so much copied anymore....did a lot of changes to it :stuck_out_tongue_closed_eyes: )

Closes #365


Alexandcoats:
Is tokio's signal not already cross-platform?

Alexandcoats:
I recommend creating a `docker` feature which would compile this code differently, because this does not suit shutdown outside of docker.

bin/inx-chronicle/src/shutdown.rs
```diff
@@ -0,0 +1,63 @@
...
+        } else {
+            shutdown_procedure(shutdown_tx);
+        }
+    });
+
+    shutdown_rx
+}
+
+fn shutdown_procedure(shutdown_tx: ShutdownTx) {
+    warn!("Gracefully shutting down the node, this may take some time.");
```
grtlr, line 58:
```suggestion
    warn!("Gracefully shutting down the Chronicle, this may take some time.");
```

Alex6323: without the `the` I'll take it :trollface:

shufps:
Works perfectly :partying_face: 
```
testnet root@fat-hornet-0():/opt/chronicle# docker-compose down -t500
Stopping inx-chronicle ... done
Stopping mongo         ... done
Removing inx-chronicle ... done
Removing mongo         ... done
Network web is external, skipping
```

log output:
```
inx-chronicle    | [2022-06-30T14:29:56Z WARN  inx_chronicle::shutdown] Gracefully shutting down the node, this may take some time.
inx-chronicle    | [2022-06-30T14:29:56Z DEBUG inx_chronicle::api] API Worker shutting down (b11c9bc3-39a4-4274-ae89-3064660b3216)
inx-chronicle    | [2022-06-30T14:29:56Z INFO  inx_chronicle::api] Stopping API server
inx-chronicle    | [2022-06-30T14:29:56Z DEBUG inx_chronicle::metrics] Metrics Worker shutting down (c012787b-aef4-404e-9205-5d971788fa85)
testnet root@fat-hornet-0():/opt/chronicle# 
```

Cargo.toml
```diff
@@ -117,6 +117,9 @@ console = [
     "dep:console-subscriber",
     "tokio/tracing",
 ]
+docker = [
+    "default",
```
Alexandcoats, line 121: This should be `docker = []` and the `default` feature should contain `"docker"`.

bin/inx-chronicle/src/shutdown.rs
```diff
@@ -0,0 +1,63 @@
...
+
+pub(crate) type ShutdownTx = oneshot::Sender<()>;
+pub(crate) type ShutdownRx = oneshot::Receiver<()>;
+
+#[cfg(all(unix, feature = "docker"))]
+pub(crate) use tokio::signal::unix::SignalKind;
+
+/// Creates a shutdown listener for Unix platforms.
+#[cfg(all(unix, feature = "docker"))]
+pub(crate) fn shutdown_listener(signals: Vec<SignalKind>) -> ShutdownRx {
```
Alexandcoats, line 15: Just make these async rather than using channels?

bin/inx-chronicle/src/shutdown.rs
```diff
@@ -0,0 +1,63 @@
...
+        } else {
+            shutdown_procedure(shutdown_tx);
+        }
+    });
+
+    shutdown_rx
+}
+
+/// Creates a shutdown listener for Non-Unix platforms.
+#[cfg(not(all(unix, feature = "docker")))]
```
Alexandcoats, line 42: These should be `#[cfg(all(not(unix), feature = "docker"))]`

bin/inx-chronicle/src/main.rs
```diff
@@ -38,8 +39,19 @@ async fn main() {
 async fn startup(scope: &mut RuntimeScope) -> Result<(), Box<dyn Error + Send + Sync>> {
     let launcher_addr = scope.spawn_actor_unsupervised(Launcher).await;
 
-    spawn_task("ctrl-c listener", async move {
```
Alexandcoats, line 41: Keep this impl when the `docker` feature is not enabled.

bin/inx-chronicle/src/shutdown.rs
Alex6323: Sure? This has to be the negation of the above `#[cfg(all(unix, feature = "docker"))]` to really be complementary, and what you propose is not the logical negation.

Cargo.toml
Alex6323: Done

bin/inx-chronicle/src/shutdown.rs
Alexandcoats: You're right, I was assuming there would be a third feature set but this is the old impl.

grtlr (changes requested):
We shouldn't need the `docker` feature. I looked at `tokio::signal::ctrl_z` and it maps to `SIGINT`.

bin/inx-chronicle/src/shutdown.rs
Alex6323: Done

grtlr (dismissed):

PR #373, opened 2022-07-04
fix(db): prevent duplicate inserts of `LedgerUpdateDocument`s

Alex6323:


src/db/collections/ledger_update.rs
```diff
@@ -119,19 +119,31 @@ impl MongoDb {
...
                     address: owner,
                     output_id: metadata.output_id,
-                    at: metadata.spent.map_or(metadata.booked, |s| s.spent),
-                    is_spent: metadata.spent.is_some(),
+                    at,
+                    is_spent,
                 };
 
+                let mut doc = bson::to_document(&ledger_update_document)?;
+                doc.insert("_id", ledger_update_document.output_id.to_hex());
```
grtlr, line 135: The `outputId` is not unique in this collection, because there can be the same output id for spent and unspent outputs.

grtlr:
We should maybe add the `is_spent` field as a last field of the index. That would allow us to set the index to `unique`. At least then we would get a warning from the database if we insert duplicated documents.

grtlr (approved):

PR #385, opened 2022-07-06
feat(db): use MongoDB replicata sets

### LINKED ISSUE #381: Use MongoDB replica sets in Docker
For transactions in MongoDB we need to enable replica sets. We should update the `docker-compose` file to set this up automatically.

Here is more background information on how to do this: https://www.mongodb.com/docs/manual/tutorial/convert-standalone-to-replica-set/

We should also add a warning to the documentation that people should take time to configure there MongoDB instance.

Alex6323:
Closes #381

### DIFF
--- a/docker/docker-compose.hornet.yml
+++ b/docker/docker-compose.hornet.yml
@@ -4,7 +4,13 @@ services:
   mongo:
     image: mongo
     restart: always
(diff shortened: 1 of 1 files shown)


grtlr:
We will merge this as part of #359.

PR #388, opened 2022-07-06
feat(api): implement `balance/` endpoint

Alex6323:


src/db/collections/ledger_update.rs
```diff
@@ -262,6 +262,58 @@ impl MongoDb {
...
+            locked_balance: Amount,
+        }
+
+        let balances = self
+            .0
+            .collection::<Balances>(LedgerUpdateDocument::COLLECTION)
+            .aggregate(
+                vec![
+                    doc! { "$match": {
+                        "address": &address, "is_spent": false,
```
grtlr, line 285: I'm afraid that this is not going to be enough to get the unspent outputs. This collection contains potentially two rows per output id:

* one for the creation event and
* one for the spending event

That means initially every output is inserted with `is_spent == false`. So to check if an output is really unspent we would have to:

1. Check if `is_spent == false`
2. Check if there exists not other row with the same output id where `is_spent == true`.

At this point, I'm thinking if it would make more sense to use the newly created `OutputDocument` collection for the balance computation.

grtlr: @Alexandcoats 👀

Alex6323: You're right. I think I assumed the row would be updated

src/db/collections/block.rs
```diff
@@ -20,7 +20,7 @@ use crate::{
 
 /// Chronicle Block record.
 #[derive(Clone, Debug, Serialize, Deserialize)]
-struct BlockDocument {
+pub(crate) struct BlockDocument {
```
grtlr, line 23: Why do we need those changes?

src/db/collections/outputs.rs
```diff
@@ -14,30 +14,38 @@ use crate::{
...
 /// Chronicle Output record.
 #[derive(Clone, Debug, Serialize, Deserialize)]
-struct OutputDocument {
+pub(crate) struct OutputDocument {
     output_id: OutputId,
     output: Output,
     metadata: OutputMetadata,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    address: Option<Address>,
+    is_trivial_unlock: bool,
```
grtlr, line 30: I wonder if we should create a new struct for values that we have precomputed? I'd like to keep these things separated as much as possible.

Alex6323: But then you have to come up with a name!! :trollface:

src/db/collections/block.rs
Alex6323: We don't need them anymore since I moved the balance calc. into the outputs collection. Without it we cannot do lookups from other collections.

grtlr: Let's just change only what's necessary for now, reduces the amount of code to look at when searching for errors.

src/db/collections/outputs.rs
grtlr: How about any of these: `Output{PreCalculated|Latent}{Details|Information}`?

src/db/collections/block.rs
Alex6323: what errors? :trollface:

src/db/collections/outputs.rs
Alex6323: alright, thanks. I'll throw a die on each of the choices.

src/db/collections/outputs/mod.rs
```diff
@@ -85,6 +109,24 @@ impl MongoDb {
             )
             .await?;
 
+        collection
+            .create_index(
+                IndexModel::builder()
+                    .keys(doc! { "address": 1 })
```
Alexandcoats, line 115: Should be `details.address` now I believe.

src/db/collections/outputs/mod.rs
```diff
@@ -219,4 +261,70 @@ impl MongoDb {
...
+                .try_next()
+                .await?
+                .map(bson::from_document::<Balances>)
+                .transpose()?
+                .unwrap_or_default();
+
+            Ok(Some(BalancesResult {
+                total_balance: balances.total_balance.amount as u64,
+                sig_locked_balance: balances.sig_locked_balance.amount as u64,
+                ledger_index: ledger_index.into(),
```
Alexandcoats, line 324: Why the `.into()` here?

src/db/collections/outputs/mod.rs
Alex6323: of course ... you're right ... how could I miss that :see_no_evil:

grtlr (approved):

PR #402, opened 2022-07-08
fix(db): add `unlock_condition` to `id_index`

### LINKED ISSUE #400: `id_index` on `LedgerUpdateDocument` is not unique
Because multiple addresses can "own" an an output through the different unlock conditions. It could even be that the same address appears in the different unlock conditions of the same output.

grtlr:
Closes #400.

### DIFF
--- a/src/db/collections/ledger_update.rs
+++ b/src/db/collections/ledger_update.rs
@@ -14,7 +14,7 @@ use crate::{
     db::MongoDb,
     types::{
         ledger::{MilestoneIndexTimestamp, OutputWithMetadata},
-        stardust::block::{Address, OutputId},
+        stardust::block::{Address, OutputId, UnlockConditionDescription},
         tangle::MilestoneIndex,
     },
 };
@@ -25,6 +25,9 @@ struct LedgerUpdateDocument {
     address: Address,
     output_id: OutputId,
     at: MilestoneIndexTimestamp,
+    #[serde(with = "crate::types::util::stringify")]
+    amount: u64,
+    unlock_condition: UnlockConditionDescription,
     is_spent: bool,
 }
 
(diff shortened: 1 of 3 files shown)


src/db/collections/ledger_update.rs
```diff
@@ -140,12 +143,14 @@ impl MongoDb {
             let is_spent = metadata.spent.is_some();
 
             // Ledger updates
-            for owner in output.owning_addresses() {
+            for (address, unlock_condition) in output.owning_addresses() {
```
Alex6323, line 146: `unlock_condition_desc` is probably clearer to not confuse it with the "real" `UnlockCondition` type.

src/db/collections/ledger_update.rs
```diff
@@ -25,6 +25,9 @@ struct LedgerUpdateDocument {
     address: Address,
     output_id: OutputId,
     at: MilestoneIndexTimestamp,
+    #[serde(with = "crate::types::util::stringify")]
+    amount: u64,
+    unlock_condition: UnlockConditionDescription,
```
Alex6323, line 30: consider naming it `unlock_condition_desc`?

src/types/stardust/block/output/unlock_condition.rs
```diff
@@ -35,16 +35,44 @@ pub enum UnlockCondition {
     },
 }
 
+#[derive(Copy, Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(rename_all = "snake_case", tag = "kind")]
+pub enum UnlockConditionDescription {
```
Alex6323, line 40: I really don't like this additional type. Tbh it just bloats the code base.

Alex6323 (approved):

PR #415, opened 2022-07-12
fix(api): remove `inx` from `is_healthy` check

### LINKED ISSUE #414: Remove `inx` check from `is_healthy`
This does not play nicely with runtime configuration of features.

grtlr:
Closes #414.

### DIFF
--- a/bin/inx-chronicle/src/api/routes.rs
+++ b/bin/inx-chronicle/src/api/routes.rs
@@ -8,7 +8,7 @@ use axum::{
     routing::{get, post},
     Extension, Json, Router,
 };
-use chronicle::{db::MongoDb, runtime::ScopeView};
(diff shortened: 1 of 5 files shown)


bin/inx-chronicle/src/api/routes.rs
```diff
@@ -8,22 +8,26 @@ use axum::{
...
 };
-use chronicle::{db::MongoDb, runtime::ScopeView};
+use chronicle::db::MongoDb;
 use hyper::StatusCode;
 use serde::Deserialize;
+use time::{Duration, OffsetDateTime};
 
 use super::{auth::Auth, config::ApiData, error::ApiError, responses::*, ApiResult};
 
+const STALE_MILESTONE_DURATION: Duration = Duration::minutes(1);
```
Alex6323, line 18: Can you document this constant and why we think 1 minute is a good heuristic?

Alex6323 (dismissed):

PR #428, opened 2022-07-13
fix(ci): improve feature handling and CI

### LINKED ISSUE #179: Test feature combination in `canary.yml` CI
For example using [`cargo hack`](https://github.com/taiki-e/cargo-hack).

### LINKED ISSUE #358: Compilation fails with `--no-default-features` option
### Rust version

rustc 1.61.0 (fe5b13d68 2022-05-18)

Alex6323:
Closes #358
Closes #179

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,11 +1,11 @@
 [alias]
 ci-check-all = "check --all-targets --all-features"
 ci-check-inx = "check --all-targets --no-default-features --features inx,stardust"
-ci-check-api = "check --all-targets --no-default-features --features api-core,api-history,api-analytics,stardust"
+ci-check-api = "check --all-targets --no-default-features --features api,stardust"
 
 ci-clippy-all = "clippy --all-targets --all-features -- -D warnings"
 ci-clippy-inx = "clippy --all-targets --no-default-features --features inx,stardust -- -D warnings"
-ci-clippy-api = "clippy --all-targets --no-default-features --features api-core,api-history,api-analytics,stardust -- -D warnings"
+ci-clippy-api = "clippy --all-targets --no-default-features --features api,stardust -- -D warnings"
 
 ci-doctest = "test --doc --all-features --release"
 ci-doc = "doc --all-features --no-deps --document-private-items"
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,7 +1,7 @@
 [alias]
-ci-build-all = "build --all-targets --all-features"
-ci-build-inx = "build --all-targets --no-default-features --features inx,stardust"
-ci-build-api = "build --all-targets --no-default-features --features api-core,api-history,analytics,stardust"
+ci-check-all = "check --all-targets --all-features"
+ci-check-inx = "check --all-targets --no-default-features --features inx,stardust"
+ci-check-api = "check --all-targets --no-default-features --features api-core,api-history,analytics,stardust"
 
 ci-clippy-all = "clippy --all-targets --all-features -- -D warnings"
 ci-clippy-inx = "clippy --all-targets --no-default-features --features inx,stardust -- -D warnings"
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,11 +1,11 @@
 [alias]
 ci-check-all = "check --all-targets --all-features"
 ci-check-inx = "check --all-targets --no-default-features --features inx,stardust"
-ci-check-api = "check --all-targets --no-default-features --features api-core,api-history,analytics,stardust"
+ci-check-api = "check --all-targets --no-default-features --features api-core,api-history,api-analytics,stardust"
 
 ci-clippy-all = "clippy --all-targets --all-features -- -D warnings"
 ci-clippy-inx = "clippy --all-targets --no-default-features --features inx,stardust -- -D warnings"
-ci-clippy-api = "clippy --all-targets --no-default-features --features api-core,api-history,analytics,stardust -- -D warnings"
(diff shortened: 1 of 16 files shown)


.cargo/config.toml
```diff
@@ -1,7 +1,7 @@
 [alias]
-ci-build-all = "build --all-targets --all-features"
-ci-build-inx = "build --all-targets --no-default-features --features inx,stardust"
-ci-build-api = "build --all-targets --no-default-features --features api-core,api-history,analytics,stardust"
+ci-check-all = "check --all-targets --all-features"
+ci-check-inx = "check --all-targets --no-default-features --features inx,stardust"
+ci-check-api = "check --all-targets --no-default-features --features api-core,api-history,analytics,stardust"
```
grtlr, line 4: If we do this, we also need to remove the `--release` flags from the tests (which we probably should do anyways).

Alex6323: Actually no. If we'ld do that then the tests would run significantly slower.

grtlr:
Can you also please unify all `api-*` features under a single `api`?

.cargo/config.toml
Alex6323: Why do you think the above change should also mean removing `--release` from `ci-test`? If you think it would reuse the `build` it didn't because previously it was a `debug` build anyway, not a `release` build, so `ci-test` compiled an extra release build.

grtlr: You're right—I got confused there. The thing is right now the `--release` build is probably more expansive than what we gain from running the tests a little bit faster, which is why we should probably remove the `--release` from the tests for the time being.

bin/inx-chronicle/src/api/routes.rs
```diff
@@ -62,31 +62,36 @@ async fn login(
...
+        match database
+            .get_sync_data(start.milestone_index..=end.milestone_index)
+            .await
+        {
+            Ok(sync) => sync.gaps.is_empty(),
+            _ => false,
+        }
     }
+    #[cfg(not(feature = "stardust"))]
+    true
```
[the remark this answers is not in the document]
Alex6323: I'll have a look!

bin/inx-chronicle/src/api/routes.rs
```diff
@@ -61,51 +61,59 @@ async fn login(
...
+    if is_healthy(&database).await {
         StatusCode::OK
     } else {
         StatusCode::SERVICE_UNAVAILABLE
     }
 }
 
-#[cfg(feature = "api-history")]
+#[cfg(all(feature = "stardust", feature = "api"))]
 pub async fn sync(database: Extension<MongoDb>) -> ApiResult<SyncDataDto> {
```
Alexandcoats, line 125: This could be moved to the history folder, since that's where it's used.

bin/inx-chronicle/src/api/stardust/mod.rs
```diff
@@ -14,19 +11,12 @@ pub fn routes() -> Router {
     #[allow(unused_mut)]
     let mut router = Router::new();
 
-    #[cfg(feature = "analytics")]
+    #[cfg(feature = "api")]
```
Alexandcoats, line 14: Shouldn't be needed here, as the whole api folder is gated.

src/types/mod.rs
```diff
@@ -4,11 +4,13 @@
 #![allow(missing_docs)] // TODO Remove this once everything has settled.
 
 /// Module containing the ledger data models.
+#[cfg(feature = "stardust")]
```
Alexandcoats, line 7: Maybe we should do some folder reorg instead of these weird gates.

bin/inx-chronicle/src/api/routes.rs
```diff
@@ -61,51 +61,59 @@ async fn login(
...
-    if is_healthy(database).await {
+    if is_healthy(&database).await {
         StatusCode::OK
     } else {
         StatusCode::SERVICE_UNAVAILABLE
     }
 }
 
-#[cfg(feature = "api-history")]
+#[cfg(all(feature = "stardust", feature = "api"))]
```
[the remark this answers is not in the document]
Alex6323: yea, it's a left-over. Removed!

src/db/collections/ledger_update.rs
```diff
@@ -264,7 +264,7 @@ impl MongoDb {
     }
 }
 
-#[cfg(feature = "analytics")]
+#[cfg(feature = "api")]
```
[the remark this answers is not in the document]
Alex6323: Isn't that only necessary for the analytics api?

src/types/mod.rs
grtlr: I like that.

src/db/collections/ledger_update.rs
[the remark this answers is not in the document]
grtlr: We have many such methods in the `lib` part of `inx-chronicle`. I don't think we should feature gate them all—if we don't use them in the `bin` they won't show up in the final binary anyways.

src/types/mod.rs
Alex6323: Let's do this in another PR. I'll open an issue for it. Fine with that?

bin/inx-chronicle/src/api/routes.rs
Alex6323: Done

src/types/mod.rs
grtlr: Yes, fine for me.

Alex6323: Created #437

src/db/collections/ledger_update.rs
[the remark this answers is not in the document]
Alex6323: right

grtlr (approved):

PR #459, opened 2022-07-25
fix(types): inputs commitment conversion

### LINKED ISSUE #458: `inputsCommitment`  in `blocks/{blockId}` yields different value from Hornet
When using the `blocks/{blockId}` endpoint on the same  `blockId`, the result from Chronicle differs from the result of Hornet.

Alexandcoats:
Closes #458.


grtlr (commented):
Would it take much work to write a test case for this?

Alexandcoats:
> Would it take much work to write a test case for this?

There's not really any way we can test the validity without database access. I don't think there's any need though, since the core issue is that I had incorrectly built the value when I should have just copied it.

grtlr (approved):

PR #468, opened 2022-07-25
fix(api): add max page size and tests

### LINKED ISSUE #440: Add upper bound for page sizes in APIs
Otherwise an attacker could, for example, send huge queries for addresses that are frequently used (like the faucet).

Alexandcoats:
Closes #440

### DIFF
--- a/bin/inx-chronicle/src/api/error.rs
+++ b/bin/inx-chronicle/src/api/error.rs
@@ -7,7 +7,7 @@ use axum::{
     extract::rejection::{ExtensionRejection, QueryRejection, TypedHeaderRejection},
     response::IntoResponse,
 };
-use chronicle::runtime::ErrorLevel;
+use chronicle::{db::collections::ParseSortError, runtime::ErrorLevel};
 use hyper::{header::InvalidHeaderValue, StatusCode};
 use mongodb::bson::document::ValueAccessError;
 use serde::Serialize;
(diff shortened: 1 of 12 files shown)


bin/inx-chronicle/src/api/error.rs
```diff
@@ -109,17 +109,17 @@ pub enum ParseError {
...
     BadPagingState,
-    #[error("Invalid sort order descriptor")]
-    BadSortDescriptor,
     #[cfg(feature = "stardust")]
     #[error(transparent)]
     BeeBlockStardust(#[from] bee_block_stardust::Error),
     #[error(transparent)]
     Bool(#[from] ParseBoolError),
     #[error(transparent)]
+    DecimalU256(#[from] uint::FromDecStrErr),
```
grtlr, line 118: Love the alphabetic ordering!

src/db/collections/ledger_update.rs
```diff
@@ -50,12 +53,35 @@ pub struct LedgerUpdateByMilestoneRecord {
...
 }
 
+impl Default for SortOrder {
+    fn default() -> Self {
+        Self::Newest
+    }
+}
+
+#[derive(Debug, Error)]
+#[error("Invalid sort order descriptor. Expected 'oldest' or 'newest', found '{0}'")]
```
grtlr, line 69:
```suggestion
#[error("Invalid sort order descriptor. Expected `oldest` or `newest`, found `{0}`")]
```

grtlr (approved):

PR #470, opened 2022-07-26
feat(inx): use `bee-inx`

grtlr:


bin/inx-chronicle/src/stardust_inx/cone_stream.rs
```diff
@@ -69,21 +61,19 @@ impl Actor for ConeStream {
...
-        let block_metadata = block_metadata_result?;
-        log::trace!("Block data: {:?}", block_metadata);
-        let inx_block_with_metadata: inx::BlockWithMetadata = block_metadata.try_into()?;
-        let BlockWithMetadata { metadata, block, raw } = inx_block_with_metadata;
+        let bee_inx::BlockWithMetadata { block, metadata } = block_metadata_result?;
+        log::trace!("Block id: {:?}", metadata.block_id);
 
         self.db
-            .insert_block_with_metadata(block.into(), raw, metadata.into())
+            .insert_block_with_metadata(block.clone().inner()?.into(), block.data(), metadata.into())
```
Alexandcoats, line 76: Can we not destructure this, or just pass the `BlockWithMetadata`?

grtlr: Then we would have to import `bee-inx` in the lib part of Chronicle, which I don't like. What do you mean by destructuring here?

Alexandcoats: I thought `Block` could be divided into `inner` and `data` to avoid the clone, but I took a look at `Raw` and see that is not the case.

Alexandcoats (approved):

Alexandcoats (approved):

PR #482, opened 2022-07-27
feat(analytics): implement ledger and most activity-based analytics

### LINKED ISSUE #328: Add current ledger status statistics

### LINKED ISSUE #424: Total number of tokens locked for storage deposit
Add an `/api/analytics/v2/storage-deposit` endpoint that works as follows:

```sh
curl "[...]/api/analytics/v2/storage-deposit" | jq
```

The response should look something like this:
```json
{
  "totalAmount": 142976891337,
  "sdrucAmount": 13787999,
}
```

Ping @msarcev.

### LINKED ISSUE #447: Rework analytics query to use `OutputDocument` collection
The duplicate entries in `LedgerUpdateDocument` make it very error prone to do analytics. We should consider moving the analytics to the `OutputDocument` collection.

We might want to think about augmenting the `OutputDocument` with some further convenience fields, such as `amount` and `isTrivialUnlock`.

Alexandcoats:
* Closes #447 
* Closes #328
* Closes #424


src/types/stardust/block/output/native_token.rs
```diff
@@ -95,17 +95,31 @@ impl TryFrom<TokenScheme> for bee::TokenScheme {
...
+        let amount = NativeTokenAmount::from(value.amount());
+        let (high, low) = (
+            u128::from_be_bytes(amount.0[..16].try_into().unwrap()),
+            u128::from_be_bytes(amount.0[16..].try_into().unwrap()),
+        );
         Self {
             token_id: NativeTokenId(**value.token_id()),
-            amount: value.amount().into(),
+            amount,
+            float_amount: ((high as f64) * 2f64.powi(16)) + low as f64,
```
grtlr, line 122: I don't think we should change the data model for this right now. I'm keeping a backlog of breaking database changes in the [Shimmer RC tracking issue](#486). Maybe you could add an item to the list there?

src/db/collections/outputs/mod.rs
```diff
@@ -329,3 +332,285 @@ impl MongoDb {
...
+            .await?
+            .try_next()
+            .await?
+            .map(bson::from_document)
+            .transpose()?
+            .unwrap_or_default())
+    }
+
+    /// Gathers locked storage deposit analytics.
+    pub async fn get_locked_storage_deposit_analytics(
```
grtlr, line 501: We should only compute that for the current milestone. Also, we need to sum the `OutputDetails::num_key_bytes` and `OutputDetails::num_data_bytes` of each output. Once we have those fields we can compute the byte cost by looking at the parameters in `StatusDocument`.

src/types/stardust/block/output/native_token.rs
Alexandcoats: If we don't change the data model, we cannot provide these values at all.

src/db/collections/outputs/mod.rs
Alexandcoats: What do you mean by only computing it for the current milestone?

src/db/collections/block.rs
```diff
@@ -292,49 +295,101 @@ impl MongoDb {
...
+        &self,
+        start_timestamp: Option<MilestoneTimestamp>,
+        end_timestamp: Option<MilestoneTimestamp>,
+    ) -> Result<BlockAnalyticsResult, Error> {
+        let start_index = if let Some(start_timestamp) = start_timestamp {
+            if let Some(index) = self
+                .find_first_milestone(start_timestamp)
                 .await?
-                .try_next()
+                .map(|ts| ts.milestone_index)
```
grtlr, line 347: We should probably create a helper for this in the `db` module as part of the `impl MongoDb {...}`.

Now that I think of it, would it be possible to perform this as a step in the MongoDb aggregation pipeline? If so we could create an issue for that.

src/db/collections/outputs/mod.rs
```diff
@@ -365,3 +368,419 @@ impl MongoDb {
         }
     }
 }
+
+#[derive(Clone, Debug, Default, Serialize, Deserialize)]
+pub struct OutputAnalyticsResult {
+    pub count: u64,
+    pub total_value: String,
+}
```
grtlr, line 376: Analogous to the `BlocksAnalytics`, I would love to see a combined response here:

```rust
pub struct Amount(String)

pub struct OutputStatistics {
    pub count: Amount,
    pub total_amount: Amount,
}

struct OutputsAnalyticsReponseForLackOfBetterName {
    pub combined: OutputAnalytics,
    pub alias_outputs: OutputAnalytics,
    pub basic_outputs: OutputAnalytics,
    // ...
}
```

src/db/collections/outputs/mod.rs
```diff
@@ -365,3 +368,419 @@ impl MongoDb {
...
+                                        { "metadata.booked.milestone_timestamp": { "$gte": end_timestamp } },
+                                    ] },
+                                    { "$nor": [
+                                        { "metadata.spent_metadata.spent.milestone_timestamp": { "$lt": start_timestamp } },
+                                        { "metadata.spent_metadata.spent.milestone_timestamp": { "$gte": end_timestamp } },
+                                    ] },
+                                ],
+                            } },
+                            { "$group" : { "_id": "$details.address" }},
+                            { "$count": "addresses" },
```
grtlr, line 749: Aren't we counting addresses multiple times here? I would expect some form of `distinct` or `addToSet` call here.

grtlr (changes requested):

bin/inx-chronicle/src/api/stardust/analytics/responses.rs
```diff
@@ -1,28 +1,70 @@
...
 
 /// Response of `GET /api/analytics/addresses[?start_timestamp=<i64>&end_timestamp=<i64>]`.
-#[derive(Copy, Clone, Debug, Serialize, Deserialize)]
+#[derive(Clone, Debug, Serialize, Deserialize)]
 #[serde(rename_all = "camelCase")]
 pub struct AddressAnalyticsResponse {
-    pub total_addresses: u64,
-    pub receiving_addresses: u64,
-    pub sending_addresses: u64,
+    pub total_active_addresses: String,
```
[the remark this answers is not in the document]
Alexandcoats: Those aren't amounts,.

bin/inx-chronicle/src/api/stardust/analytics/responses.rs
```diff
@@ -1,28 +1,70 @@
...
 #[derive(Copy, Clone, Debug, Serialize, Deserialize)]
 #[serde(rename_all = "camelCase")]
-pub struct TransactionsAnalyticsResponse {
-    pub count: u64,
-    pub total_value: f64,
-    pub average_value: f64,
+pub struct RentStructureDto {
+    pub v_byte_cost: u32,
+    pub v_byte_factor_data: u8,
+    pub v_byte_factor_key: u8,
```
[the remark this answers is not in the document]
Alexandcoats: I don't see one. Can you tell me where?

bin/inx-chronicle/src/api/stardust/analytics/routes.rs
```diff
@@ -4,13 +4,30 @@
...
     Router::new()
         .route("/addresses", get(address_analytics))
-        .route("/transactions", get(transaction_analytics))
+        .route("/storage-deposit", get(storage_deposit_analytics))
+        .nest(
+            "/blocks",
+            Router::new()
+                .route("/transaction", get(transaction_analytics))
+                .route("/milestone", get(milestone_analytics))
+                .route("/tagged_data", get(tagged_data_analytics)),
```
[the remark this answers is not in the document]
Alexandcoats: So you don't want the total transaction amount?

src/db/collections/block.rs
Alexandcoats: A helper for what exactly?

src/db/collections/outputs/mod.rs
Alexandcoats: This would mean creating a facet query (which would not be able to use indexes) or simply calling all of them back-to-back.

bin/inx-chronicle/src/api/stardust/analytics/responses.rs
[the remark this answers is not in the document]
grtlr: Sorry, it's called `RentStructureResponse`.

bin/inx-chronicle/src/api/stardust/analytics/routes.rs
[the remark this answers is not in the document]
grtlr: We certainly should have it, but not as part of `Block`-level analytics.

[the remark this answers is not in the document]
Alexandcoats: So you are implying that you want it in your proposed `output` level analytics?

src/db/collections/outputs/mod.rs
Alexandcoats: That's what the `$group` does?

bin/inx-chronicle/src/api/stardust/analytics/responses.rs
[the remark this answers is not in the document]
grtlr: Yeah, that name is confusing. How about `U64Response`, or something else?

bin/inx-chronicle/src/api/stardust/analytics/routes.rs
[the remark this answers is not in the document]
grtlr: Yeah, and I'd call it something like `TokenAnalytics`.

src/db/collections/block.rs
grtlr: Hmm, thinking more about it, I guess this is fine.

bin/inx-chronicle/src/api/stardust/analytics/responses.rs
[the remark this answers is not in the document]
Alexandcoats: This is just...pointless tbh

src/db/collections/mod.rs
```diff
@@ -23,3 +23,53 @@ pub use self::{
...
     treasury::TreasuryResult,
 };
+use crate::types::stardust::block::{
+    AliasOutput, BasicOutput, FoundryOutput, MilestonePayload, NftOutput, TaggedDataPayload, TransactionPayload,
+};
+
+/// Helper to specify a kind for an output type.
+pub trait OutputKind {
+    /// Gets the output kind.
+    fn kind() -> Option<&'static str> {
```
grtlr, line 33: Why do we need the `Option` here (and for `PayloadKind`)? I think we can just leave it out.

Alexandcoats: You think I would have designed it this way if I didn't have a reason? 😆 

It is so that we can _not_ specify a kind.

grtlr: Ah for the overall call?

grtlr (dismissed):

src/db/collections/mod.rs
grtlr: Gottcha

grtlr (approved):

PR #491, opened 2022-07-28
fix(db): 500 on hitting the `balance/` endpoint

Alex6323:


src/db/collections/outputs/mod.rs
```diff
@@ -320,8 +320,8 @@ impl MongoDb {
                 .unwrap_or_default();
 
             Ok(Some(BalancesResult {
-                total_balance: balances.total_balance.amount as u64,
-                sig_locked_balance: balances.sig_locked_balance.amount as u64,
+                total_balance: balances.total_balance[0].amount as u64,
```
Alexandcoats, line 323: You should be able to remove the `Balances` struct, and instead directly deserialize a `BalancesResult`, by using `$first` in a `$project`.

Alex6323: Thanks, Alex. I bet you're right on this. Will try to do it a bit later.

src/db/collections/outputs/mod.rs
```diff
@@ -81,11 +81,11 @@ pub struct OutputWithMetadataResult {
     pub metadata: OutputMetadataResult,
 }
 
-#[derive(Clone, Debug)]
+#[derive(Clone, Debug, Default, Deserialize)]
 #[allow(missing_docs)]
 pub struct BalancesResult {
-    pub total_balance: u64,
-    pub sig_locked_balance: u64,
+    pub total_balance: f64,
```
grtlr, line 87: Change this to `String` (see comment below)

grtlr (changes requested):

grtlr (changes requested):
Does the current query properly handle outputs that are spent within the same milestone? @Alexandcoats could you please also verify that?

Alex6323:
If it's spent in the same milestone then `{ "metadata.spent_metadata.spent": null },` will not match?

grtlr:
Yeah, you're right, I must have been confused by the different versions of the query.

Alexandcoats (dismissed):

Alex6323:
I'll approve as well :trollface: since @Alexandcoats made the query much simpler and possibly more performant as well.

grtlr (approved):

PR #494, opened 2022-07-29
fix(db): 500 on hitting the balance/ endpoint (alternative)

Alex6323:
This is an alternative to #491 . The differences are:
* no intermediary helper structs to deserialize the db result
* returns a balance of `0` for unknown addresses


Alex6323:
#491 was adapted after some discussion.

PR #523, opened 2022-08-04
feat(analytics): add `richest-addresses` and `token-distribution` endpoints

Alexandcoats:


bin/inx-chronicle/src/api/stardust/analytics/routes.rs
```diff
@@ -29,7 +30,8 @@ pub fn routes() -> Router {
             Router::new()
                 .route("/storage-deposit", get(storage_deposit_analytics))
                 .route("/native-tokens", get(unspent_output_analytics::<FoundryOutput>))
-                .route("/nfts", get(unspent_output_analytics::<NftOutput>)),
+                .route("/nfts", get(unspent_output_analytics::<NftOutput>))
+                .route("/richlist", get(richlist_analytics)),
```
grtlr, line 34: Nitpick: Probably something like `richest-addresses` is more appropriate? If we change it here, we should also change the other method calls.

docs/api-analytics.yml
```diff
@@ -76,6 +76,26 @@ paths:
           $ref: "#/components/responses/NoResults"
         "500":
           $ref: "#/components/responses/InternalError"
+  /api/analytics/v2/ledger/richlist:
+    get:
+      tags:
+        - ledger
+      summary: Returns a richlist.
```
grtlr, line 83: Reminder.

bin/inx-chronicle/src/api/stardust/analytics/responses.rs
```diff
@@ -50,3 +54,46 @@ pub struct StorageDepositAnalyticsResponse {
...
 
 impl_success_response!(StorageDepositAnalyticsResponse);
+
+/// Response of `GET /api/analytics/richlist[?top=<usize>]`.
+#[derive(Clone, Debug, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct RichlistAnalyticsResponse {
+    pub distribution: Vec<DistributionStatDto>,
+    pub top: Vec<AddressStatDto>,
+}
```
grtlr, line 64: I think we should split these two queries, following the logic of the other endpoints. One should be `richest-adresses` and the other one should be `token-distribution`.

Additionally, I think we should return the `ledger_index` here too.

src/db/collections/outputs/mod.rs
```diff
@@ -675,3 +675,97 @@ impl MongoDb {
...
+}
+
+impl MongoDb {
+    /// Create richlist statistics.
+    pub async fn get_richlist_analytics(
+        &self,
+        ledger_index: Option<MilestoneIndex>,
+        top: usize,
+    ) -> Result<Richlist, Error> {
+        let ledger_index = match ledger_index {
```
grtlr, line 710: `.or_else`?

bin/inx-chronicle/src/api/stardust/analytics/routes.rs
Alexandcoats: I dunno, it's more than that. I think richlist is appropriate.

docs/api-analytics.yml
Alexandcoats: Reminder for what?

bin/inx-chronicle/src/api/stardust/analytics/responses.rs
Alexandcoats: Fair enough.

src/db/collections/outputs/mod.rs
Alexandcoats: Not in _my_ async await

docs/api-analytics.yml
grtlr: Renaming `richlist` to `richest-addresses` :trollface:

src/db/collections/outputs/mod.rs
grtlr: Dang, yeah.

grtlr (approved):

PR #537, opened 2022-08-08
feat(api): add routes endpoint

Alexandcoats:


grtlr:
For now we should just go with a simple solution that adds the top level APIs to the list of routes. We should be consistent with what Hornet is doing here, because people might match on these.

PR #554, opened 2022-08-15
feat(metrics): use `tracing` instead of `log`

Alexandcoats:


Alexandcoats:
This isn't quite ready but I would still like some eyes on it.

src/db/collections/milestone.rs
```diff
@@ -169,6 +170,7 @@ impl MongoDb {
     }
 
     /// Inserts the information of a milestone into the database.
+    #[instrument(skip(self, session, payload))]
```
grtlr, line 173: I think we can skip everything except the `milestone_index`.

grtlr (commented):
LGTM

grtlr (approved):

Issue #564, opened 2022-08-17
Use MongoDB `insert_many` for unspent outputs


Alexandcoats:
Given that our data is now always time linear, we can assume that unspent outputs do not exist in the database, thus we can use `insertMany` for those.

PR #566, opened 2022-08-17
feat(db): use `insertMany` for initial unspent outputs

### LINKED ISSUE #561: Use batch inserts for unspent outputs on initial run
When the database is still empty we can use batch inserts (`insert_many`) for the unspent outputs.

Alexandcoats:
Closes #561.


Alexandcoats:
Closes #564

grtlr (commented):
Comments from our call.

src/db/collections/block.rs
```diff
@@ -210,41 +210,37 @@ impl MongoDb {
...
+        .await?;
+        let blocks_with_metadata = blocks_with_metadata
+            .into_iter()
+            .map(|block_document| {
+                let block_id = block_document.metadata.block_id;
+                let mut doc = bson::to_document(&block_document)?;
+                doc.insert("_id", block_id.to_hex());
+                Result::<_, Error>::Ok(doc)
+            })
+            .collect::<Result<Vec<_>, _>>()?;
```
grtlr, line 236: Remove

src/db/collections/block.rs
```diff
@@ -210,41 +210,37 @@ impl MongoDb {
...
 
-            self.db
-                .collection::<bson::Document>(BlockDocument::COLLECTION)
-                .insert_many_with_session(blocks_with_metadata, None, session)
+        for batch in blocks_with_metadata.chunks(10000) {
```
grtlr, line 238: Constant

src/db/collections/ledger_update.rs
```diff
@@ -120,36 +120,50 @@ impl MongoDb {
...
+                        output_id: o.metadata.output_id,
+                        at: o.metadata.spent_metadata.map(|s| s.spent).unwrap_or(o.metadata.booked),
+                        is_spent: o.metadata.spent_metadata.is_some(),
+                    })
+                } else {
+                    None
+                }
+            })
+            .collect::<Vec<_>>();
+        for batch in ledger_updates.chunks(10000) {
```
grtlr, line 139: Constant

src/db/mongodb.rs
```diff
@@ -100,6 +110,44 @@ impl MongoDb {
...
 }
 
+pub struct MongoDbCollection<T>(mongodb::Collection<T>);
+
+impl<T: Serialize> MongoDbCollection<T> {
+    /// Inserts many records and ignores duplicate key errors.
+    pub async fn insert_many_ignore_duplicates(
+        &self,
+        docs: impl IntoIterator<Item = impl Borrow<T>>,
+        options: impl Into<Option<InsertManyOptions>>,
```
grtlr, line 120: I think we should bake the unordered in instead of making it a parameter, the function does not really make sense, otherwise.

grtlr (changes requested):

src/db/mongodb.rs
```diff
@@ -100,6 +110,44 @@ impl MongoDb {
...
+    }
+}
+
+impl<T> std::ops::Deref for MongoDbCollection<T> {
+    type Target = mongodb::Collection<T>;
+
+    fn deref(&self) -> &Self::Target {
+        &self.0
+    }
+}
```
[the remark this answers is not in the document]
Alexandcoats: Well, if used properly we would use it in place of the current `collection` fn, so yes.

bin/inx-chronicle/src/stardust_inx/mod.rs
```diff
@@ -172,7 +192,8 @@ impl Actor for InxWorker {
 #[derive(Debug)]
 pub struct LedgerUpdateRecord {
     milestone_index: MilestoneIndex,
-    outputs: Vec<OutputWithMetadata>,
+    created: Vec<OutputWithMetadata>,
+    consumed: Vec<OutputWithMetadata>,
```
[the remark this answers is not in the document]
Alexandcoats: I looked into it, and it would require creating two types just to coerce them into one again for the document. Seems unnecessary.

src/db/mongodb.rs
Alexandcoats: It can still make sense. The unordered just specifies that we won't stop as soon as we hit an error.

grtlr: But if the insert is ordered Mongo will stop after hitting the first error. In which case I think it's bad to ignore the error.

Alexandcoats: I think it makes the most sense to mimic the way we call the normal function in every way.

grtlr (approved):

PR #567, opened 2022-08-17
feat(db): add some basic db tests

### LINKED ISSUE #607: Add db tests

### LINKED ISSUE #614: Prevent CI from running twice for `pull_request` and `push`
This happens only for branches in the [inx-chronicle](https://github.com/iotaledger/inx-chronicle) repository.

Alexandcoats:
Closes #607.
Closes #614.


grtlr (commented):
Please revert the formatting of the `Cargo.toml` files to single line dependencies.

Also the tests should not live in `/src` but rather in a top-level folder like `integration_tests` or just `tests`.

grtlr:
Can you please also ensure that the `coverage.yml` CI workflow picks up these tests? I'd imagine we'd have to update the script in `.github/workflows/scripts/coverage.sh`.

Alexandcoats:
> Can you please also ensure that the `coverage.yml` CI workflow picks up these tests? I'd imagine we'd have to update the script in `.github/workflows/scripts/coverage.sh`.

Actually, since that uses `cargo test --all` it should already pick up the new crate in the workspace.

grtlr:
You're right, they are run already

.gitignore
```diff
@@ -2,7 +2,7 @@
 
 **/.env
 
-/bin/inx-chronicle/config.toml
+**/config.toml
```
grtlr, line 5: Why have this line at all? I would say this is a bit to general.

Cargo.toml
```diff
@@ -32,68 +32,64 @@ dyn-clone = { version = "1.0", default-features = false }
...
 
-[dev-dependencies]
-bee-block-stardust = { package = "bee-block", version = "1.0.0-beta.5", default-features = false, features = [ "dto", "rand", "std", "serde", ] }
-packable = { version = "0.5", default-features = false }
-
 [features]
-default = [
-    "api",
-    "inx",
+default = [
```
grtlr, line 87: Why the spaces?

Cargo.toml
```diff
@@ -111,21 +107,21 @@ api = [
...
+    "dep:bee-block-stardust", 
     "dep:bee-api-types-stardust",
 ]
 
 [profile.production]
 inherits = "release"
 codegen-units = 1
 lto = true
-strip = "symbols"
+strip = "symbols"
```
grtlr, line 127: Newline

.gitignore
Alexandcoats: Why? I want to have a local config file without pushing it...

Cargo.toml
grtlr: Ah, my diff made it look like you added them. nvm

src/types/stardust/mod.rs
```diff
@@ -5,3 +5,6 @@
 pub mod block;
 /// Module containing the milestone model.
 pub mod milestone;
+
+/// Module containing utilities for stardust types.
+pub mod util;
```
grtlr, line 10: These utility functions should live with the integration tests in the `common` module.

.gitignore
grtlr: But then we should remove the stars. This removes all `config.toml` in all potential directories because of that I think it's too general.

grtlr (approved):

PR #580, opened 2022-08-22
fix(inx): use `insert_many` on initial sync

grtlr:


grtlr:
Closing in favor of #564.

PR #585, opened 2022-08-23
fix(api): return array of single output id for indexer responses

### LINKED ISSUE #584: Fix indexer foundry ID route not returning array
[Indexer route `/api/indexer/v1/outputs/foundry/{foundryId}` returns an array containing the output ID](https://editor.swagger.io/?url=https://raw.githubusercontent.com/iotaledger/tips/indexer-api/tips/TIP-0026/indexer-rest-api.yaml) but Chronicle returns the output id directly as the value.

expected:
```
{
  ledgerIndex: 622037,
  pageSize: 0,
  items: [
    '0x8439fc69c98e36290315b4472e64409ceb2757804399b0edf5b3727ed0f41fe60100'
  ]
}
```

got:
```
{
  ledgerIndex: 622055,
  outputId: '0x8439fc69c98e36290315b4472e64409ceb2757804399b0edf5b3727ed0f41fe60100'
}
```

Alexandcoats:
Fixes #584

### DIFF
--- a/bin/inx-chronicle/src/api/stardust/indexer/responses.rs
+++ b/bin/inx-chronicle/src/api/stardust/indexer/responses.rs
@@ -9,7 +9,7 @@ use crate::api::responses::impl_success_response;
 #[serde(rename_all = "camelCase")]
 pub struct IndexerOutputResponse {
     pub ledger_index: u32,
-    pub output_id: String,
+    pub items: Vec<String>,
 }
 
 impl_success_response!(IndexerOutputResponse);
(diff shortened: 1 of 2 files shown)


bin/inx-chronicle/src/api/stardust/indexer/responses.rs
```diff
@@ -9,7 +9,7 @@ use crate::api::responses::impl_success_response;
 #[serde(rename_all = "camelCase")]
 pub struct IndexerOutputResponse {
     pub ledger_index: u32,
-    pub output_id: String,
+    pub items: Vec<String>,
 }
```
grtlr, line 13: We should be able to just reuse the `IndexerOutputsResponse` from below. The TIP actually uses the same schema for both:

image

grtlr (changes requested):

grtlr (approved):

PR #596, opened 2022-08-25
feat(db)!: proper use of `_id` fields

### LINKED ISSUE #568: Use `#[serde(rename = "_id")]` field where appropriate
This should reduce the amount of indexes and would guarantee that our insert logic works.

Collections that come to mind are:

* `MilestoneDocument::milestone_id`, so that we don't have to touch as many queries)
* `BlockDocument::block_id`
* `OutputDocument::output_id` (more tricky, because it's currently only in the metadata)

### LINKED ISSUE #598: Duplicate key error on `insert_unspent_outputs`
Some of this might be solved by not calling `insert_unspent_outputs` and `insert_spent_outputs` concurrently. Also, and this might be related, these lines look off:

https://github.com/iotaledger/inx-chronicle/blob/87839eeca58d1b8e5c785a88ff5b84fb7ae12075/src/db/mongodb.rs#L115-L126

Why are we always returning `Ok(0)` when there were errors? Shouldn't this depend on the ordering? Maybe we should return a richer struct (`inserted`, `skipped`)?

What happens if we have multiple write errors? Right now, it seems like we are only checking on `write_errs.len() == 1` but couldn't there be multiple errors at once?

grtlr:
Closes #568.
Closes #598.

### DIFF
--- a/src/db/collections/outputs/mod.rs
+++ b/src/db/collections/outputs/mod.rs
@@ -31,6 +31,8 @@ use crate::{
 /// Chronicle Output record.
 #[derive(Clone, Debug, Serialize, Deserialize)]
 struct OutputDocument {
+    #[serde(rename = "_id")]
+    output_id: OutputId,
     output: Output,
     metadata: OutputMetadata,
     details: OutputDetails,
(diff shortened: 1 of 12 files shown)


src/db/collections/outputs/mod.rs
```diff
@@ -175,9 +162,9 @@ impl MongoDb {
         self.db
             .collection::<OutputDocument>(OutputDocument::COLLECTION)
             .update_one(
-                doc! { "metadata.output_id": output_id },
-                doc! { "$set": bson::to_document(&OutputDocument::from(output))? },
-                UpdateOptions::builder().upsert(true).build(),
+                doc! { "_id": output_id },
```
Alexandcoats, line 165: Not sure if it's okay to do this, because we will end up with dangling metadata with no underlying output data if we never inserted the unspent piece.

Alexandcoats (dismissed):

PR #601, opened 2022-08-26
feat(api): allow configuring argon for JWT

Alexandcoats:


grtlr (commented):
I would not expose all the argon stuff in the `config.toml`. We should either

* provide sub-command `inx-chronicle jwt`, or
* automatically generate the JWT in a dedicated output folder.

For both options, we  should choose sensible defaults where possible.

Alexandcoats:
> I would not expose all the argon stuff in the `config.toml`. We should either
> 
>     * provide sub-command `inx-chronicle jwt`, or
> 
>     * automatically generate the JWT in a dedicated output folder.
> 
> 
> For both options, we should choose sensible defaults where possible.

I don't understand. This is fundamentally JWT configuration, so it should go there.

grtlr (approved):

Alex6323 (approved):

PR #604, opened 2022-08-29
feat(db): use `db.run_command` for faster bulk updates

### LINKED ISSUE #605: Improve performance of inserting spent outputs
We currently use `replace_one` to update outputs that are spent, which is prohibitively slow for larger ledger updates.

grtlr:
Closes #605.


Alexandcoats:
Can we extract this to a fn in `MongoDb`?

PR #620, opened 2022-08-30
fix!(api): return Bech32s in `richest-addresses` endpoint

grtlr:
Closes: #578.

### DIFF
--- a/src/bin/inx-chronicle/api/stardust/analytics/routes.rs
+++ b/src/bin/inx-chronicle/api/stardust/analytics/routes.rs
@@ -177,11 +177,9 @@ async fn richest_addresses_ledger_analytics(
         .ok_or(ApiError::NoResults)?;
 
     let hrp = database
-        .get_latest_protocol_parameters()
+        .get_protocol_parameters_for_ledger_index(res.ledger_index)
         .await?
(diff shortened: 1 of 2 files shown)


src/bin/inx-chronicle/api/stardust/analytics/routes.rs
```diff
@@ -175,8 +176,24 @@ async fn richest_addresses_ledger_analytics(
         .await?
         .ok_or(ApiError::NoResults)?;
 
+    let hrp = database
+        .get_latest_protocol_parameters()
+        .await?
+        .ok_or(ApiError::Internal(InternalApiError::CorruptState(
```
Alexandcoats, line 182: No need to manually wrap the `InternalApiError` as there is a `From` definition.

src/bin/inx-chronicle/api/stardust/analytics/routes.rs
```diff
@@ -175,8 +176,24 @@ async fn richest_addresses_ledger_analytics(
         .await?
         .ok_or(ApiError::NoResults)?;
 
+    let hrp = database
+        .get_latest_protocol_parameters()
```
Alexandcoats, line 180: Shouldn't we use the ledger index to retrieve these?

Alexandcoats (approved):

PR #626, opened 2022-08-31
feat(db): separate database collections into individual types

Alexandcoats:
This PR creates individual types for each DB collection. In order to make it as clean as possible, I created a new API trait `MongoDbCollection` to enable calling genericised functions and convert the underlying collection types.

These collections have a small quirk: Some of their methods need to access more than one collection. For this reason, some of them hold multiple mongodb `Collections`. There is probably a better way to do this.

Additionally, I added a thread-safe version of the chunks iterator from `itertools` which mostly works. Sometimes it causes inexplicable errors in the wrong locations, but I believe these are actually compiler bugs and not real issues. Either way, there is only one location where it was necessary to work around this: inserting treasury payloads.

I have observed a significant improvement in syncing times with these changes, due to the reduction of cloning.

### DIFF
--- a/src/bin/inx-chronicle/stardust_inx/chunks.rs
+++ b/src/bin/inx-chronicle/stardust_inx/chunks.rs
@@ -3,7 +3,7 @@
 
 use std::sync::{
     atomic::{AtomicUsize, Ordering},
-    Arc, Mutex,
+    Mutex,
 };
 
 pub trait ChunksExt: Iterator {
@@ -12,7 +12,7 @@ pub trait ChunksExt: Iterator {
         Self: Sized,
     {
         IntoChunks {
-            inner: Arc::new(Mutex::new(GroupInner {
+            inner: Mutex::new(GroupInner {
                 key: ChunkIndex::new(size),
                 iter: self,
                 current_key: None,
@@ -23,7 +23,7 @@ pub trait ChunksExt: Iterator {
                 bottom_group: 0,
                 buffer: Vec::new(),
                 dropped_group: !0,
-            })),
+            }),
             index: AtomicUsize::new(0),
         }
     }
@@ -31,7 +31,7 @@ pub trait ChunksExt: Iterator {
 impl<T: Iterator> ChunksExt for T {}
 
 pub struct IntoChunks<I: Iterator> {
-    inner: Arc<Mutex<GroupInner<usize, I, ChunkIndex>>>,
+    inner: Mutex<GroupInner<usize, I, ChunkIndex>>,
     index: AtomicUsize,
 }
 
@@ -66,8 +66,7 @@ impl<'a, I: Iterator> Iterator for Chunks<'a, I> {
     #[inline]
     fn next(&mut self) -> Option<Self::Item> {
         let index = self.parent.index.fetch_add(1, Ordering::Relaxed);
-        let inner = &mut *self.parent.inner.lock().unwrap();
-        inner.step(index).map(|elt| Chunk {
+        self.parent.step(index).map(|elt| Chunk {
             parent: self.parent,
             index,
             first: Some(elt),
--- a/src/bin/inx-chronicle/stardust_inx/chunks.rs
+++ b/src/bin/inx-chronicle/stardust_inx/chunks.rs
@@ -1,0 +1,345 @@
+// Copyright 2022 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+use std::sync::{
+    atomic::{AtomicUsize, Ordering},
+    Arc, Mutex,
+};
+
+pub trait ChunksExt: Iterator {
+    fn chunks(self, size: usize) -> IntoChunks<Self>
+    where
+        Self: Sized,
+    {
+        IntoChunks {
+            inner: Arc::new(Mutex::new(GroupInner {
+                key: ChunkIndex::new(size),
+                iter: self,
+                current_key: None,
+                current_elt: None,
+                done: false,
+                top_group: 0,
+                oldest_buffered_group: 0,
+                bottom_group: 0,
+                buffer: Vec::new(),
+                dropped_group: !0,
+            })),
+            index: AtomicUsize::new(0),
+        }
+    }
+}
+impl<T: Iterator> ChunksExt for T {}
+
+pub struct IntoChunks<I: Iterator> {
+    inner: Arc<Mutex<GroupInner<usize, I, ChunkIndex>>>,
+    index: AtomicUsize,
+}
+
+impl<I: Iterator> IntoChunks<I> {
+    /// `client`: Index of chunk that requests next element
+    fn step(&self, client: usize) -> Option<I::Item> {
+        self.inner.lock().unwrap().step(client)
+    }
+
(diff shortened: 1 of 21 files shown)


src/bin/inx-chronicle/stardust_inx/chunks.rs
```diff
@@ -0,0 +1,344 @@
...
+    F: for<'a> KeyFunction<&'a I::Item, Key = K>,
+    K: PartialEq,
+{
+    /// `client`: Index of group that requests next element
+    #[inline(always)]
+    fn step(&mut self, client: usize) -> Option<I::Item> {
+        // println!("client={}, bottom_group={}, oldest_buffered_group={}, top_group={}, buffers=[{}]",
+        // client, self.bottom_group, self.oldest_buffered_group,
+        // self.top_group,
+        // self.buffer.iter().map(|elt| elt.len()).format(", "));
```
Alex6323, line 136: wanna still keep that for debugging?

Alexandcoats: Yeah, just in case.

src/bin/inx-chronicle/stardust_inx/chunks.rs
```diff
@@ -0,0 +1,344 @@
+// Copyright 2022 IOTA Stiftung
```
Alex6323, line 1: You maybe want to also port and adapt the unit tests from `itertools`? (also ... maybe consider giving tribute to the original authors - I haven't checked against their code though)

Alexandcoats: There are no tests to adapt. And regarding crediting them, well, I don't think there's any reason to do that. This isn't a library 😆

src/bin/inx-chronicle/stardust_inx/chunks.rs
```diff
@@ -0,0 +1,344 @@
...
+    fn drop(&mut self) {
+        self.parent.drop_group(self.index);
+    }
+}
+
+impl<'a, I: Iterator> Iterator for Chunk<'a, I> {
+    type Item = I::Item;
+    #[inline]
+    fn next(&mut self) -> Option<Self::Item> {
+        if let elt @ Some(..) = self.first.take() {
```
Alex6323, line 93: ugh, that's ugly, haha. Isn't this basically the same as `if let Some(elt) = self.first.take()`?

src/bin/inx-chronicle/stardust_inx/chunks.rs
```diff
@@ -0,0 +1,344 @@
...
+        if elt.is_none() && client == self.oldest_buffered_group {
+            // FIXME: VecDeque is unfortunately not zero allocation when empty,
+            // so we do this job manually.
+            // `bottom_group..oldest_buffered_group` is unused, and if it's large enough, erase it.
+            self.oldest_buffered_group += 1;
+            // skip forward further empty queues too
+            while self
+                .buffer
+                .get(self.oldest_buffered_group - self.bottom_group)
+                .map_or(false, |buf| buf.len() == 0)
```
Alex6323, line 169: Isn't clippy complaining here about not using `buf.is_empty()` instead?

src/bin/inx-chronicle/stardust_inx/chunks.rs
Alexandcoats: No, it binds `elt` to the `Option` not the interior value.

src/bin/inx-chronicle/stardust_inx/chunks.rs
Alexandcoats: It isn't for me. Is it for you?

src/bin/inx-chronicle/stardust_inx/chunks.rs
Alex6323: ah, alright, now I see it!

src/bin/inx-chronicle/stardust_inx/chunks.rs
Alex6323: I dunno, I haven't tried to compile this PR locally, but it's unlikely that it does because the CI is also happy with it. Seems like Clippy isn't capable of seeing through this in all cases.

Alex6323: Or maybe they removed the lint.

src/bin/inx-chronicle/stardust_inx/chunks.rs
```diff
@@ -0,0 +1,344 @@
...
+            match self.current_key.take() {
+                None => {}
+                Some(old_key) => {
+                    if old_key != key {
+                        self.current_key = Some(key);
+                        first_elt = Some(elt);
+                        break;
+                    }
+                }
+            }
```
Alex6323, line 230: I think `if let Some(old_key) = self.current_key.take() {` should do it.

Alex6323: Probably use `matches!` with the `if clause` inside like you did elsewhere.

src/bin/inx-chronicle/stardust_inx/chunks.rs
Alex6323: maybe it's just me who finds this code non-trivial :see_no_evil:

src/bin/inx-chronicle/stardust_inx/mod.rs
```diff
@@ -252,19 +273,43 @@ impl HandleEvent<Result<LedgerUpdateRecord, InxError>> for InxWorker {
...
-        db.insert_unspent_ledger_updates(outputs.iter())
-    )?;
+    let output_collection = db.collection::<OutputCollection>();
+    let ledger_collection = db.collection::<LedgerUpdateCollection>();
+    try_join! {
+        async {
+            for batch in &outputs.iter().chunks(INSERT_BATCH_SIZE) {
+                output_collection.insert_unspent_outputs(batch).await?;
+            }
+            Result::<_, InxError>::Ok(())
```
Alex6323, line 283: The compiler can't resolve the exact Result, but for the other async block it can?

Alex6323 (approved):

src/bin/inx-chronicle/stardust_inx/chunks.rs
Alex6323: yea

src/bin/inx-chronicle/stardust_inx/mod.rs
Alexandcoats: Yes, because of this one.

src/bin/inx-chronicle/stardust_inx/chunks.rs
```diff
@@ -0,0 +1,395 @@
...
+    #[inline]
+    fn next(&mut self) -> Option<Self::Item> {
+        if let elt @ Some(..) = self.first.take() {
+            return elt;
+        }
+        self.parent.step(self.index)
+    }
+}
+
+struct GroupInner<K, I, F>
```
Alex6323, line 101: Is a group semantically the same as a chunk?

PR #650, opened 2022-09-12
feat(db)!: separate database collections into individual types (#626)

### LINKED ISSUE #594: Create structs for each of the collections in `db`
This should help avoid redundant calls like `self.db.collection::<BlockDocument>(BlockDocument::COLLECTION)`.

Only the structs required for a given query should be passed as `Extension`s to the API. That way we can easily see which API uses which collection.

Ideally, these structs should also hide as much of the implementation details (like the `*Document` structs) as possible.

grtlr:
* Separate database collections into individual types.

* Create ext traits to simplify usage.

* Use custom thread safe iterator for chunking.

* Write chunks tests and cleanup code.

Closes #594.


src/db/mongodb.rs
```diff
@@ -175,3 +140,125 @@ impl Default for MongoDbConfig {
...
+
+    fn collection(&self) -> &mongodb::Collection<Self::Document>;
+
+    fn with_type<T>(&self) -> mongodb::Collection<T> {
+        self.collection().clone_with_type()
+    }
+}
+
+#[async_trait]
+pub trait MongoCollectionExt: MongoDbCollection {
```
grtlr, line 160: `MongoDbCollectionExt`

PR #676, opened 2022-09-13
chore(test): separate test lib

Alexandcoats:


grtlr (commented):
Having the random generation of types in a separate crate (`bee-test`) was kind of a pain in Bee. A much cleaner solution is to put the random generation under a `rand` feature flag.

Then we add something like this for each type:

```rust
#[cfg(any(feature = "test", feature = "rand"))]
impl Block {
    pub fn random() -> Self {
        // ...
    }
}
```

Then, the integration tests can enable that feature. The main advantages of doing things this way are:

* Locality, things that belong together are close together in code
* Discoverability, it works nicely with auto-complete
* Less `use ...` statements in the tests

Cargo.toml
```diff
@@ -120,6 +120,9 @@ opentelemetry = [
     "dep:opentelemetry-jaeger",
     "dep:tracing-opentelemetry",
 ]
+rand = [
+    "bee-block-stardust?/rand"
```
grtlr, line 124: Why can this just be `dep:bee-block-stardust/rand"`?

src/types/stardust/block/address/mod.rs
```diff
@@ -66,29 +66,47 @@ impl From<Address> for Bson {
...
+        }
+
+        /// Generates a ed25519 [`Address`].
+        pub fn rand_ed25519() -> Self {
+            Self::Ed25519(Ed25519Address::rand())
+        }
+    }
+}
+
+#[cfg(all(test, feature = "rand"))]
```
grtlr, line 91: I think here we should only have `#[cfg(test)]`, whereas `#[cfg(any(test, feature = "rand"))]` should go on `mod rand`.

_Notice the `any` in the `cfg` statement above._

I think in it's current form, we would not be able to run any tests without the `rand` feature.

src/types/stardust/block/address/alias.rs
```diff
@@ -39,3 +39,17 @@ impl From<AliasAddress> for Bson {
         mongodb::bson::to_bson(&val).unwrap()
     }
 }
+
+#[cfg(feature = "rand")]
```
grtlr, line 43: These files don't have `#[cfg(any(test, ...))]` because they don't contain unit tests, right?

src/types/stardust/block/input.rs
```diff
@@ -35,20 +35,51 @@ impl TryFrom<Input> for bee::Input {
...
+        }
+
+        /// Generates a random treasury [`Input`].
+        pub fn rand_treasury() -> Self {
+            Self::from(&bee::Input::from(rand_treasury_input()))
+        }
+    }
+}
+
+#[cfg(all(test, feature = "rand"))]
```
grtlr, line 70: Same here

src/types/stardust/block/output/alias.rs
```diff
@@ -142,24 +143,45 @@ impl TryFrom<AliasOutput> for bee::AliasOutput {
...
+
+    impl AliasOutput {
+        /// Generates a random [`AliasOutput`].
+        pub fn rand() -> Self {
+            rand_alias_output().into()
+        }
+    }
+}
+
+#[cfg(all(test, feature = "rand"))]
```
grtlr, line 167: Same here

src/types/stardust/block/address/mod.rs
Alexandcoats: > I think in it's current form, we would not be able to run any tests without the rand feature.

Yes, because they all depend on random data. That's kind of the point. This way we can write tests that don't depend on random data without the extra feature gate.

src/types/stardust/block/mod.rs
```diff
@@ -62,25 +62,94 @@ impl TryFrom<Block> for bee::BlockDto {
...
+                protocol_version: rand_number(),
+                parents: BlockId::rand_parents(),
+                payload: None,
+                nonce: rand_number(),
+            }
+        }
+    }
+}
+
+#[cfg(all(test, feature = "rand"))]
```
grtlr, line 135:
```suggestion
```

PR #677, opened 2022-09-13
feat(db): separate ledger/protocol_param logic from collections

Alexandcoats:


src/bin/inx-chronicle/api/stardust/analytics/routes.rs
```diff
@@ -114,6 +116,15 @@ async fn unspent_output_ledger_analytics<O: OutputKind>(
...
 ) -> ApiResult<OutputAnalyticsResponse> {
+    let ledger_index = if let Some(ledger_index) = ledger_index {
+        ledger_index
+    } else {
+        database
+            .collection::<MilestoneCollection>()
+            .get_ledger_index()
+            .await?
+            .ok_or(ApiError::NoResults)?
+    };
```
grtlr, line 127: Can we move this to a separate function, like `async fn or_newest(db: &MongoDb, ledger_index: Option<MilestoneIndex>`?

grtlr (dismissed):

grtlr:
I want to merge #678 first, to not delay it any further.

src/bin/inx-chronicle/api/stardust/analytics/routes.rs
Alexandcoats: Honestly, I don't think that would help at all.

grtlr: I had it like that when I experimented with it and I think it really helped with descriptiveness. It also removes some of the redundancy in the code.

src/bin/inx-chronicle/api/stardust/core/routes.rs
```diff
@@ -250,19 +253,24 @@ fn create_output_metadata_response(metadata: OutputMetadataResult) -> OutputMeta
             .map(|spent_md| spent_md.transaction_id.to_hex()),
         milestone_index_booked: *metadata.booked.milestone_index,
         milestone_timestamp_booked: *metadata.booked.milestone_timestamp,
-        ledger_index: metadata.ledger_index.0,
+        ledger_index: ledger_index.0,
```
grtlr, line 256: I'd use the `metadata` here. If for some reason, something is wrong with the `ledger_index` it will be a lot harder to debug.

Alexandcoats: It's literally the same value. There's no reason to push it through the db call to to pull it back.

grtlr (approved):

PR #692, opened 2022-09-19
feat: add `claiming` endpoint to `anlytics/v2`

### LINKED ISSUE #682: Provide endpoint for claiming information
It would be great to provide information about the claiming process. We're mainly interested in how many tokens have been claimed.

I think there are two relevant statistics that we could help provide:

* How many tokens have been claimed overall so far?
* How many tokens have been claimed per milestone?

Here is how we can get the information that we need:

An `Output` was part of the genesis if its corresponding `BlockId` is `0x0`, or if its booked `MilestoneIndex` is `0`. Then, all we have to do is check if it has been spent.

grtlr:
Closes #682.

### DIFF
--- a/src/bin/inx-chronicle/api/stardust/analytics/routes.rs
+++ b/src/bin/inx-chronicle/api/stardust/analytics/routes.rs
@@ -23,8 +23,9 @@ use super::{
     extractors::{LedgerIndex, MilestoneRange, RichestAddressesQuery},
     responses::{
         ActivityPerInclusionStateDto, ActivityPerPayloadTypeDto, AddressAnalyticsResponse, AddressStatDto,
-        MilestoneAnalyticsResponse, OutputAnalyticsResponse, OutputDiffAnalyticsResponse, RichestAddressesResponse,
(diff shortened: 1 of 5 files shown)


src/bin/inx-chronicle/api/stardust/analytics/routes.rs
```diff
@@ -44,6 +45,7 @@ pub fn routes() -> Router {
             "/activity",
             Router::new()
                 .route("/addresses", get(address_activity_analytics))
+                .route("/claiming", get(claimed_tokens_analytics))
```
Alexandcoats, line 48: Shouldn't it be `claimed-tokens`?

Alexandcoats (dismissed):

src/bin/inx-chronicle/api/stardust/analytics/routes.rs
grtlr: Yeah, you're right. :pray:

Alexandcoats (approved):

PR #698, opened 2022-09-19
feat(db): use a materialized view for ledger updates

### LINKED ISSUE #686: Turn `stardust_ledger_updates` into an on-demand materialized view

Alexandcoats:
Closes #686

### DIFF
--- a/src/bin/inx-chronicle/stardust_inx/mod.rs
+++ b/src/bin/inx-chronicle/stardust_inx/mod.rs
@@ -11,8 +11,7 @@ use bee_inx::client::Inx;
 use chronicle::{
     db::{
         collections::{
-            BlockCollection, LedgerUpdateCollection, MilestoneCollection, OutputCollection, ProtocolUpdateCollection,
-            TreasuryCollection,
+            BlockCollection, MilestoneCollection, OutputCollection, ProtocolUpdateCollection, TreasuryCollection,
         },
         MongoDb,
     },
@@ -24,7 +23,6 @@ use chronicle::{
     },
 };
 use futures::{StreamExt, TryStreamExt};
-use tokio::try_join;
 use tracing::{debug, info, instrument, trace_span, warn, Instrument};
 
 use self::{chunks::ChunksExt, stream::LedgerUpdateStream};
@@ -187,6 +185,8 @@ impl Actor for InxWorker {
                 task.await.unwrap()?;
             }
 
+            self.db.collection::<OutputCollection>().create_ledger_updates().await?;
+
             info!("Inserted {} unspent outputs.", count);
 
             info!(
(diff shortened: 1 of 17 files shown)


src/db/collections/outputs/mod.rs
```diff
@@ -213,6 +214,68 @@ impl OutputCollection {
...
+                    "_id": {
+                        "milestone_index": { "$ifNull": [
+                            "$metadata.spent_metadata.spent.milestone_index",
+                            "$metadata.booked.milestone_index"
+                        ] },
+                        "output_id": "$_id",
+                        "is_spent": { "$ne": [ "$metadata.spent_metadata", null ] },
+                    },
+                    "address": "$details.address",
+                    "milestone_timestamp": "$metadata.booked.milestone_timestamp",
```
Alex6323, line 267: Does that always have to be the `timestamp` of the "booking" milestone?

Alexandcoats: Ah, good catch.

grtlr: Good catch, I think you're right. This should probably use the `$ifNull` mechanism too.

@Alexandcoats Could you please at this to your test case as well?

src/db/collections/outputs/mod.rs
```diff
@@ -213,6 +214,74 @@ impl OutputCollection {
...
+                        ] },
+                        "output_id": "$_id",
+                        "is_spent": { "$ne": [ "$metadata.spent_metadata", null ] },
+                    },
+                    "address": "$details.address",
+                    "milestone_timestamp": { "$ifNull": [
+                        "$metadata.spent_metadata.spent.milestone_timestamp",
+                        "$metadata.booked.milestone_timestamp"
+                    ] },
+                } },
```
grtlr, line 237: As the `$project` stage is identical I would love to pull this out into a function to reduce the redundancy.

Alexandcoats: Let's worry about this as part of our refactoring of these later.

src/bin/inx-chronicle/stardust_inx/mod.rs
```diff
@@ -187,6 +185,8 @@ impl Actor for InxWorker {
                 task.await.unwrap()?;
             }
 
+            self.db.collection::<OutputCollection>().create_ledger_updates().await?;
```
grtlr, line 188:
```suggestion
            self.db.collection::<OutputCollection>().create_ledger_updates_view().await?;
```

src/bin/inx-chronicle/stardust_inx/mod.rs
```diff
@@ -256,6 +256,10 @@ impl HandleEvent<Result<LedgerUpdateRecord, InxError>> for InxWorker {
 
         insert_unspent_outputs(&self.db, ledger_update.created).await?;
         update_spent_outputs(&self.db, ledger_update.consumed).await?;
+        self.db
+            .collection::<OutputCollection>()
+            .merge_into_ledger_updates(ledger_update.milestone_index)
```
grtlr, line 261:
```suggestion
            .merge_into_ledger_updates_view(ledger_update.milestone_index)
```

src/db/collections/outputs/mod.rs
Alexandcoats: I mean pulling out the pipeline stages into functions or whatever.

grtlr: We should grab these low-hanging fruits though, especially if we have already identified areas for improvement.

Alexandcoats: Unfortunately it is not low hanging in my opinion. The right solution is not as clear as simply pulling this into a function. I think we need to give it some thought and apply a uniform solution everywhere.

grtlr (approved):

PR #699, opened 2022-09-20
feat(test): add more db query tests

Alex6323:
# Notes for the reviewer

* [x] Run tests;


grtlr:
@Alex6323 Can you please fix the CI so that we can merge it?

src/db/collections/block.rs
```diff
@@ -94,6 +94,16 @@ impl BlockCollection {
         Ok(())
     }
 
+    // /// Gets the length of the collection.
```
Alexandcoats, line 97: Remove?

grtlr: Yes, please remove this and rebase. It would be great to get this merged before the end of this sprint.

tests/blocks.rs
```diff
@@ -113,22 +112,135 @@ mod test_rand {
...
+        assert_eq!(
+            collection
+                .get_spending_transaction(&output_id)
+                .await
+                .unwrap()
+                .map(|b| b.parents),
+            Some(block.parents)
+        );
+        // assert_eq!(collection.get_spending_transaction(&output_id).await.unwrap().map(|b| b.payload),
+        // Some(block.payload));
```
grtlr, line 223: What's up with this?

grtlr: Can we open an issue for it, so that it is not blocking this PR? Maybe add a TODO with the link to the issue to the file as well.

Alex6323: it's fixed now by limiting the spending transaction to only have one output, but I can open an issue so we can fix this later. The test is valid though as it proves what it's supposed to even if there's only 1 output.

src/types/stardust/block/payload/transaction.rs
```diff
@@ -253,6 +253,34 @@ mod rand {
                 },
             }
         }
+
+        /// Generates a random [`TransactionEssence`].
+        pub fn rand_spending(
```
Alexandcoats, line 258: What's the purpose for this extra fn?

tests/ledger_updates.rs
```diff
@@ -0,0 +1,243 @@
...
+                },
+                output,
+                output_id,
+                rent_structure: RentStructureBytes {
+                    num_key_bytes: 0,
+                    num_data_bytes: 100,
+                },
+            })
+            .inspect(|booked_output| {
+                assert!(!booked_outputs.contains(booked_output));
```
Alexandcoats, line 204: So there are two options here, either this can realistically fail or it cannot. If it can, then we should be using login to ensure that they are unique. If not, then remove this check.

tests/ledger_updates.rs
```diff
@@ -0,0 +1,243 @@
...
+                booked_outputs.push(booked_output.clone());
+            })
+            .enumerate()
+            .filter_map(|(i, booked_output)| {
```
Alexandcoats, line 208: Feels like you should just create the booked list, then iterate and `filter_map` to created the spent list.

src/types/stardust/block/payload/transaction.rs
Alex6323: `rand` possibly creates inputs and outputs that I don't want in my particular test-case scenario. I also need access to those inputs and outputs. For all created inputs `get_spending_transaction` has to return the same block. Since unspent outputs are not stored with the block, I need to separatedly insert them into the Outputcollection, and finally check that fetched block's payload (which is re-assembled through a `$lookup`) matches the one from the originally created random block.

tests/ledger_updates.rs
```diff
@@ -0,0 +1,243 @@
...
+        assert_eq!(booked_outputs.len(), 100);
+        assert_eq!(spent_outputs.len(), 50);
+
+        collection
+            .insert_unspent_ledger_updates(booked_outputs.iter())
+            .await
+            .unwrap();
+
+        collection
+            .insert_spent_ledger_updates(spent_outputs.iter())
```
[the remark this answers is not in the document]
Alex6323: But this is more from the global tangle/ledger perspective ... you cannot spent an output that wasn't unspent BEFORE.

src/types/stardust/block/payload/transaction.rs
Alex6323: fixed

tests/ledger_updates.rs
Alex6323: I think that would be better, yeah.

tests/common/mod.rs
```diff
@@ -32,5 +32,18 @@ pub async fn connect_to_test_db(database_name: impl ToString) -> Result<MongoDb,
...
     config.database_name = database_name.to_string();
 
-    Ok(MongoDb::connect(&config).await?)
+    let db = MongoDb::connect(&config).await?;
+    db.clear().await?;
+    Ok(db)
+}
+
+#[allow(unused)]
+pub async fn setup_coll<T: MongoDbCollection + Send + Sync>(db: &MongoDb) -> Result<T, TestDbError> {
```
Alexandcoats, line 41: Can we please use the full names when it's not extremely long?

tests/ledger_updates.rs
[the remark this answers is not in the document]
Alexandcoats: Well, that's actually not true for us. We insert spent outputs in each milestone before the unspent ones.

tests/ledger_updates.rs
```diff
@@ -175,15 +175,13 @@ mod test_rand {
     }
 
     #[tokio::test]
-    async fn test_spent_unspent_ledger_updates() {
+    async fn test_insert_spent_ledger_updates() {
         let db = setup_db("test-spent-unspent-ledger-updates").await.unwrap();
```
Alexandcoats, line 179: Missed the db name

Alex6323: thanks

tests/common/mod.rs
Alex6323: Done

PR #717, opened 2022-09-22
fix(db): create indexes on `.milestone_index`

### LINKED ISSUE #714: Output indexes should be on `MilestoneIndex` instead of `MilestoneIndexTimestamp`
When we do this change, we should also rename the indexes so that users don't have to delete and recreate indexes.

grtlr:
Closes #714.

### DIFF
--- a/src/db/collections/outputs/indexer/mod.rs
+++ b/src/db/collections/outputs/indexer/mod.rs
@@ -392,10 +392,10 @@ impl OutputCollection {
 
         self.create_index(
             IndexModel::builder()
-                .keys(doc! { "metadata.booked": -1 })
+                .keys(doc! { "metadata.booked.milestone_index": -1 })
                 .options(
                     IndexOptions::builder()
-                        .name("output_booked_index".to_string())
+                        .name("output_booked_milestone_index".to_string())
                         .partial_filter_expression(doc! { "metadata.spent_metadata": null })
                         .build(),
                 )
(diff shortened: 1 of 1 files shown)


Alex6323 (dismissed):

src/db/collections/outputs/indexer/mod.rs
```diff
@@ -392,10 +392,10 @@ impl OutputCollection {
 
         self.create_index(
             IndexModel::builder()
-                .keys(doc! { "metadata.booked": -1 })
+                .keys(doc! { "metadata.booked.milestone_index": -1 })
                 .options(
                     IndexOptions::builder()
-                        .name("output_booked_index".to_string())
+                        .name("output_booked_milestone_index".to_string())
                         .partial_filter_expression(doc! { "metadata.spent_metadata": null })
```
Alexandcoats, line 399: We should really also remove this filter. None of our queries that use this index specify it.

src/db/collections/outputs/indexer/mod.rs
```diff
@@ -406,11 +406,11 @@ impl OutputCollection {
...
         self.create_index(
             IndexModel::builder()
-                .keys(doc! { "metadata.spent_metadata.spent": -1 })
+                .keys(doc! { "metadata.spent_metadata.spent.milestone_index": -1 })
                 .options(
                     IndexOptions::builder()
-                        .name("output_spent_index".to_string())
-                        .partial_filter_expression(doc! { "metadata.spent_metadata.spent": { "$exists": true } })
+                        .name("output_spent_milestone_index".to_string())
+                        .partial_filter_expression(doc! { "metadata.spent_metadata": { "$exists": true } })
```
Alexandcoats, line 413: I believe this must specify down to the milestone index field.

Alexandcoats (approved):

Alex6323 (approved):
lgtm

PR #719, opened 2022-09-22
feat(db): add document count to `MongoDbCollectionExt` trait

Alex6323:


src/db/mongodb/collection.rs
```diff
@@ -111,6 +112,13 @@ pub trait MongoDbCollectionExt: MongoDbCollection {
     ) -> Result<UpdateResult, Error> {
         self.with_type().replace_one(query, replacement, options).await
     }
+
+    /// Returns the number of documents in the collection.
+    #[cfg(feature = "rand")]
+    async fn len(&self) -> Result<usize, Error> {
```
grtlr, line 118: I would probably call the method `.count()` to avoid the clippy warning and to signal that this is potentially a more expensive computation.

src/db/mongodb/collection.rs
```diff
@@ -111,6 +112,13 @@ pub trait MongoDbCollectionExt: MongoDbCollection {
     ) -> Result<UpdateResult, Error> {
         self.with_type().replace_one(query, replacement, options).await
     }
+
+    /// Returns the number of documents in the collection.
+    #[cfg(feature = "rand")]
```
grtlr, line 117: Maybe we can leave out the feature gate here.

src/db/mongodb/collection.rs
Alex6323: fine with me

src/db/mongodb/collection.rs
Alex6323: We could, but atm we have no other use-case for that functionality except for checking correctness of collection inserts.

Alex6323: alright

grtlr (approved):

PR #725, opened 2022-09-22
feat(cli)!: rework CLI args usage

Alexandcoats:
# Notes for the reviewer

* [x] Run Chronicle with various command-line options including `--help` and subcommands;


src/bin/inx-chronicle/cli.rs
```diff
@@ -1,30 +1,35 @@
...
     /// The location of the configuration file.
-    #[clap(short, long)]
+    #[clap(short, long, env = "CONFIG_PATH")]
     pub config: Option<String>,
-    /// The address of the MongoDB database.
-    #[clap(long = "db")]
-    pub db_addr: Option<String>,
+    /// The MongoDB connection string.
+    #[clap(long, env = "MONGODB_CONN_URI")]
+    pub mongodb_conn_url: Option<String>,
```
grtlr, line 17:
```suggestion
    pub mongodb_url: Option<String>,
```

src/bin/inx-chronicle/cli.rs
```diff
@@ -33,4 +38,97 @@ pub struct ClArgs {
...
+                            uuid::Uuid::new_v4().to_string(),
+                            ApiData::AUDIENCE,
+                        )
+                        .unwrap()
+                        .expires_after_duration(api_data.jwt_expiration)
+                        .expect("invalid JWT config"),
+                        api_data.secret_key.as_ref(),
+                    )
+                    .expect("invalid JWT config");
+                    println!("Bearer {}", jwt);
```
grtlr, line 119: I think it would be a nice touch to also print out the information to the user for when this token will expire.

src/bin/inx-chronicle/cli.rs
Alexandcoats: Any reasoning? I think it's good to specify that this isn't simply the address, but the connection string.

Alexandcoats: I am fine with that. @grtlr ?

grtlr: I'm sorry I just now learned that there is more than just an URL to this. I'm fine with `mongodb_conn_str`.

grtlr:
Can you rebase please?

src/bin/inx-chronicle/cli.rs
```diff
@@ -33,4 +38,97 @@ pub struct ClArgs {
...
+                #[cfg(feature = "api")]
+                Subcommands::GenerateJWT => {
+                    use crate::api::ApiData;
+                    let api_data = ApiData::try_from(config.api.clone()).expect("invalid API config");
+                    let claims = auth_helper::jwt::Claims::new(
+                        ApiData::ISSUER,
+                        uuid::Uuid::new_v4().to_string(),
+                        ApiData::AUDIENCE,
+                    )
+                    .unwrap()
```
grtlr, line 112: undocumented `unwrap`

src/bin/inx-chronicle/cli.rs
```diff
@@ -33,4 +38,97 @@ pub struct ClArgs {
...
+        if let Some(enabled) = self.enable_metrics {
+            config.metrics.enabled = enabled;
+        }
+
+        Ok(config)
+    }
+
+    /// Process subcommands and return whether the app should early exit.
+    #[allow(unused)]
+    #[allow(clippy::collapsible_match)]
```
grtlr, line 99: Is there really no way around ignoring these lints?

src/bin/inx-chronicle/cli.rs
```diff
@@ -33,4 +38,97 @@ pub struct ClArgs {
...
+    pub fn process_subcommands(&self, config: &ChronicleConfig) -> bool {
```
grtlr, line 100: Shouldn't this return an error type?

grtlr (changes requested):

src/bin/inx-chronicle/cli.rs
Alexandcoats: Not that I can tell.

src/bin/inx-chronicle/cli.rs
Alexandcoats: I don't think it's necessary for CLI operations.

grtlr: What if there was an error during any of the subcommands?

Alexandcoats: Then it will panic 😆

grtlr: Panics should be reserved for things that we didn't anticipate to happen. Since some of these commands can very well return errors we should have a mechanism to catch them and shut down things properly. This will also be required once we have more administrative functionality in the CLI.

Alexandcoats: This doesn't start anything up to shut down.

grtlr:
This should be a breaking change and signal that we have changed the config template.

src/bin/inx-chronicle/cli.rs
```diff
@@ -33,4 +41,98 @@ pub struct ClArgs {
...
+                    .expires_after_duration(api_data.jwt_expiration)
+                    .map_err(crate::api::ApiError::InvalidJwt)?;
+                    let exp_ts = time::OffsetDateTime::from_unix_timestamp(claims.exp.unwrap() as _).unwrap();
+                    let jwt = auth_helper::jwt::JsonWebToken::new(claims, api_data.secret_key.as_ref())
+                        .map_err(crate::api::ApiError::InvalidJwt)?;
+                    println!("Bearer {}", jwt);
+                    println!("Expires: {}", exp_ts);
+                    return Ok(true);
+                }
+                _ => (),
```
grtlr, line 126: Now that we have error types, can't we return something like `unknown subcommand` here?

grtlr (commented):
This is probably unrelated, but if I do `cargo run --release -- --help`, I get the following error:

`error: Found an argument which wasn't expected or isn't valid in this context`

Could you please look into this as well?

docker/docker-compose.yml
```diff
@@ -28,12 +28,12 @@ services:
       - RUST_LOG=warn,inx_chronicle=debug
     tty: true
     command:
-      - "--inx"
+      - "-c"
+      - "config.toml"
+      - "--inx-addr"
       - "http://hornet:9029"
-      - "--db"
+      - "--mongodb-conn-url"
```
Alex6323, line 35: `mongodb-conn-str` for consistency?

src/bin/inx-chronicle/cli.rs
```diff
@@ -1,30 +1,38 @@
...
+    /// The MongoDB connection string.
+    #[clap(long, env = "MONGODB_CONN_STR")]
+    pub mongodb_conn_str: Option<String>,
     /// The address of the INX interface provided by the node.
-    #[clap(long = "inx", env = "INX_ADDR")]
+    #[clap(long, env = "INX_ADDR")]
+    #[cfg(feature = "inx")]
     pub inx_addr: Option<String>,
-    /// Toggle INX (offline mode).
+    /// Toggle INX write workflow.
```
Alex6323, line 25: by "write workflow" you actually mean "import"?

Alex6323 (dismissed):
lgtm

grtlr (approved):

Alex6323 (approved):

PR #732, opened 2022-09-24
doc(wiki): fix broken links

Dr-Electron:
Fixed the core API links in the docs


Dr-Electron:
You can undo that commit again :trollface: . The problem is the imported README which references the CHANGELOG. But the docs can’t resolve that relative link.

PR #744, opened 2022-09-27
fix(db): add index on `metadata.block_id`

### LINKED ISSUE #691: Create index on `metadata.block_id` for Outputs
This would make the `$lookup` for Outputs belonging to a block faster. It would also allow us the reference the `block_id` in #682.

grtlr:
Closes #691.


Alex6323 (approved):
that might come in handy /s

Alexandcoats (approved):

PR #748, opened 2022-09-27
doc(wiki): remove version info

Dr-Electron:
With wiki V3 we can get rid of this. Chronicle docs for Shimmer will be hosted under /shimmer/chronicle/...


grtlr:
Thank you!

grtlr (approved):

PR #752, opened 2022-09-28
chore(db)!: direct dto conversion

### LINKED ISSUE #733: Convert types directly to Bee's DTOs
Right now, we get retrieve our types from MongoDB. We then convert these types to Bee types and then use Bee's conversion to create Bee DTOs. With the changes to `Packable`, we should think about providing `impl From` for our types to Bee DTOs directly, to avoid having to fetch the `ProtocolParameters` from the database for each query.

Alexandcoats:
Closes #733

### DIFF
--- a/src/types/stardust/block/payload/milestone/mod.rs
+++ b/src/types/stardust/block/payload/milestone/mod.rs
@@ -5,7 +5,7 @@ mod milestone_id;
 
 use std::borrow::Borrow;
 
-use bee_block_stardust::payload::milestone as bee;
+use bee_block_stardust::payload::milestone::{self as bee};
 use serde::{Deserialize, Serialize};
 
 pub use self::milestone_id::MilestoneId;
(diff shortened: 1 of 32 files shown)


src/types/stardust/block/payload/milestone/mod.rs
```diff
@@ -40,30 +40,45 @@ impl TryFromWithContext<MilestonePayload> for bee::MilestonePayload {
...
+            signatures: value.signatures.into_vec().into_iter().map(Into::into).collect(),
+        }
     }
 }
 
 #[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
 pub struct MilestoneEssence {
     pub index: MilestoneIndex,
     pub timestamp: u32,
+    pub protocol_version: u8,
```
grtlr, line 81: This change will force us to wipe the database, right?

grtlr (changes requested):
I think we will have to wipe the database with the changes to the `MilestoneEssence`, or provide some mitigation logic. Any thoughts?

Please adapt the PR title accordingly.

src/types/stardust/block/payload/milestone/mod.rs
grtlr: No worries.

src/types/stardust/block/payload/milestone/mod.rs
```diff
@@ -256,6 +316,7 @@ mod rand {
             Self {
                 index: rand_number::<u32>().into(),
                 timestamp: rand_number::<u32>(),
+                protocol_version: 0,
```
grtlr, line 319:
```suggestion
                protocol_version: rand_number::<u8>(),
```

grtlr (dismissed):

PR #756, opened 2022-09-29
chore(deps): use latest bee-inx `1.0.0-rc.1`

Alex6323:


Alex6323:
The plan now is to get rid of `bee-inx` and move its contents to Chronicle.

PR #757, opened 2022-09-29
feat(api): additional raw endpoints

### LINKED ISSUE #755: Provide remaining raw endpoints
We are currently missing some raw endpoints. I've found missing checks for the `headers` for these endpoints:

* `/transactions/:transactionID/included-block`
* `/outputs/:outputId`

According to @muXxer, the following endpoints in Hornet support raw headers:

* `/blocks/:blockID`
* `/transactions/:transactionID/included-block`
* `/milestones/:milestoneID`
* `/milestones/by-index/:milestoneIndex`
* `/outputs/:outputID`

And we should support the same.

Alex6323:
## Linked Issues

* Closes #755

## Notes to Reviewer

* [x] Ensure that the API response times scale with database size appropriately.
* [x] Check for breaking changes in the API and matching commit message.

* [x] Check for breaking changes in the data model and matching commit message.
 
### DIFF
--- a/src/db/collections/outputs/mod.rs
+++ b/src/db/collections/outputs/mod.rs
@@ -657,9 +657,9 @@ impl OutputCollection {
                     },
                     doc! { "$project": {
                         "output_count": { "$first": "$all.output_count" },
-                        "storage_deposit_return_count": { "$first": "$storage_deposit.return_count" },
+                        "storage_deposit_return_count": { "$ifNull": [ { "$first": "$storage_deposit.return_count" }, 0 ] },
                         "storage_deposit_return_total_value": { 
-                            "$toString": { "$first": "$storage_deposit.return_total_value" } 
+                            "$toString": { "$ifNull": [ { "$first": "$storage_deposit.return_total_value" }, 0 ] } 
                         },
                         "total_key_bytes": { 
                             "$toString": { "$first": "$all.total_key_bytes" } 
--- a/src/db/collections/outputs/mod.rs
+++ b/src/db/collections/outputs/mod.rs
@@ -81,7 +81,6 @@ impl From<&LedgerOutput> for OutputDocument {
     fn from(rec: &LedgerOutput) -> Self {
         let address = rec.output.owning_address().copied();
         let is_trivial_unlock = rec.output.is_trivial_unlock();
-        let rent_structure = rec.output.rent_structure();
 
         Self {
             output_id: rec.output_id,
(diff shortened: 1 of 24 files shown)


src/db/collections/outputs/mod.rs
```diff
@@ -96,6 +99,7 @@ impl From<&LedgerOutput> for OutputDocument {
                 is_trivial_unlock,
                 rent_structure,
             },
+            raw: rec.raw.clone(),
```
Alexandcoats, line 102: ❌  I don't think this is a good idea. We can just pack the outputs.

Alex6323: you mean at query time?

grtlr: I agree with @Alexandcoats

Alex6323: I agree transitively.

grtlr: 😁

src/types/stardust/block/output/mod.rs
```diff
@@ -152,6 +153,12 @@ impl Output {
             Self::Foundry(_) => true,
         }
     }
+
+    pub fn raw(self) -> Result<Vec<u8>, bee_block_stardust::Error> {
+        let ctx = bee_block_stardust::protocol::protocol_parameters();
+        let bee_output = bee_block_stardust::output::Output::try_from_with_context(&ctx, self)?;
+        Ok(bee_output.pack_to_vec())
+    }
```
grtlr, line 161: `bee_block_stardust::protocol::protocol_parameters()` represents the Shimmer protocol parameters and should mainly be used for testing. This is very fragile (it sometimes does not work with the `private_tangle` for example). We should retrieve the proper context here, if context-free conversions are not available.

grtlr (changes requested):

src/types/stardust/block/output/mod.rs
Alex6323: I know that, but for some reason I need the token supply to being able to create bee output type. I haven't grasped that fully yet. Any suggestion?

Alex6323: I look into another way ...

grtlr: Yeah, since the fields in `bee-block` are mainly private, we can't really create these types in an ad-hoc manner. The only option that we have right now is to get the protocol parameters that were active at the time the output was issued. We have been in talks with @thibault-martinez to allow such functionality directly in Bee.

grtlr:
For some reason, the result of `/outputs` is still JSON and not binary when I test it locally. @Alex6323 @Alexandcoats can you please verify as well?

Alexandcoats (approved):

grtlr (approved):

PR #760, opened 2022-09-29
feat(db): add created index stats

### LINKED ISSUE #743: Report statistics about indexes and their creation on startup
We had some issues with indexes not being generated on a first run. It would be good to look into this again. While we are added we should add a compact summary of the indexes to the logs (similar to what we do with the DB size and collections already). The following things should be dispayed:

* The number of indexes before calling `create_index`.
* The names of newly created indexes (only if there are already indexes in the DB).
* If there are no indexes just report the number.

Alexandcoats:
## Linked Issues

* Closes #743 

## Notes to Reviewer

* [x] Run a new instance of chronicle with some indexes disabled and view startup output.
* [x] Re-enable indexes and view startup output.
 

src/bin/inx-chronicle/launcher.rs
```diff
@@ -77,18 +77,37 @@ impl Actor for Launcher {
...
-                .await?;
-            db.collection::<chronicle::db::collections::MilestoneCollection>()
+            db.collection::<collections::MilestoneCollection>()
                 .create_indexes()
                 .await?;
+            let end_indexes = db.get_index_names().await?;
+            for (collection, indexes) in end_indexes {
+                if let Some(old_indexes) = start_indexes.get(&collection) {
+                    if indexes.difference(old_indexes).count() > 0 {
+                        info!("Created {} new indexes in {}", indexes.len(), collection);
```
Alex6323, line 100: You probably want the length of the difference.

src/db/mongodb/mod.rs
```diff
@@ -51,11 +53,28 @@ impl MongoDb {
...
         T::instantiate(self, self.db.collection(T::NAME))
     }
 
+    /// Gets all index names by their collection.
+    pub async fn get_index_names(&self) -> Result<HashMap<String, HashSet<String>>, Error> {
+        let mut res = HashMap::new();
+        for collection in self.db.list_collection_names(None).await? {
+            let indexes = self.db.collection::<Document>(&collection).list_index_names().await?;
+            if !indexes.is_empty() {
+                res.insert(collection, indexes.into_iter().collect());
```
Alex6323, line 72: Couldn't you just move the whole `Vec` into that `HashMap` instead of collecting it into a `HashSet`?

Alex6323: ah...I see you want to keep the lookup performance.

Alexandcoats: Well, yes, but part of the check uses `HashSet::difference` so either way I need that.

Alex6323 (approved):

src/bin/inx-chronicle/launcher.rs
```diff
@@ -77,18 +77,38 @@ impl Actor for Launcher {
 
         #[cfg(feature = "stardust")]
         {
-            db.collection::<chronicle::db::collections::OutputCollection>()
+            use chronicle::db::collections;
+            let start_indexes = db.get_index_names().await?;
```
Alex6323, line 81: One additional question: is it necessary that `get_index_names()` comes before the `create_collection` calls? Otherwise it would probably make sense to move it right before the index creation logic.

Alexandcoats: I believe that it is necessary, because creating the collections creates `_id` indexes.

grtlr (approved):

PR #763, opened 2022-10-03
feat(db): Re-add ledger update materialized view impl

Alexandcoats:
I have re-added the materialized view for ledger updates. However, I cannot confirm whether or not this is actually a performance improvement. Locally, I tested this against `main` with 225 milestones at 500 bps, and observed a total time of 8m 39s for this method and 9m 55s for `main`. Critically, this PR depends on #760 due to the need for indexes in the materialized view queries.

## Notes to Reviewer

* [ ] Check for breaking changes in the data model and matching commit message.
 

Alex6323 (approved):
Lgtm!

src/db/collections/outputs/mod.rs
```diff
@@ -226,6 +227,91 @@ impl OutputCollection {
...
+                } },
+                doc! { "$merge": { "into": LedgerUpdateCollection::NAME, "whenMatched": "keepExisting" } },
+            ],
+            None,
+        )
+        .await?
+        .try_next()
+        .await?;
+        Ok(())
+    }
```
Alex6323, line 313: Regarding conciseness ... I think it could be considered to merge those two into one method, and pass either a dynamic or a const generics argument to handle the two cases.

Alex6323 (approved):

Alexandcoats:
After further testing using the metrics setup, I don't think we can merge this. Even with indexes functioning properly, it is ~20% slower than the current implementation.

PR #769, opened 2022-10-05
feat: improve MongoDb connection string handling

grtlr:
## Notes to Reviewer

**Config Changes**
* [ ] Ensure proper order in which CLI and config arguments are applied.

### DIFF
--- a/src/db/mongodb/mod.rs
+++ b/src/db/mongodb/mod.rs
@@ -35,12 +35,14 @@ impl MongoDb {
         client_options.app_name = Some("Chronicle".to_string());
         client_options.min_pool_size = config.min_pool_size;
 
-        if let (Some(username), Some(password)) = (&config.username, &config.password) {
-            let credential = Credential::builder()
-                .username(username.clone())
-                .password(password.clone())
-                .build();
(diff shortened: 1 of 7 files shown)


src/db/mongodb/mod.rs
```diff
@@ -35,12 +35,14 @@ impl MongoDb {
...
-                .build();
-            client_options.credential = Some(credential);
+        if client_options.credential.is_none() {
+            if let (Some(username), Some(password)) = (&config.username, &config.password) {
+                let credential = Credential::builder()
+                    .username(username.clone())
+                    .password(password.clone())
+                    .build();
+                client_options.credential = Some(credential);
+            }
```
Alex6323, line 45: Should we error or at least log something if both didn't provide any credentials?

Alex6323: hmm, probably not necessary as the error will be detailed enough I guess.

Alex6323 (approved):
lgtm

src/db/mongodb/mod.rs
grtlr: Yeah, I'd think so.

PR #777, opened 2022-10-06
refactor: remove actor framework

### LINKED ISSUE #593: Investigate simplifying the actor framework

### LINKED ISSUE #688: Remove the actor system to get rid of `LedgerUpdateStream`

grtlr:
## Linked Issues

* Closes #593.
* Closes #688.

## Notes to Reviewer

#### INX Changes
* [ ] Ensure that the database is populated correctly.


src/bin/inx-chronicle/api/mod.rs
```diff
@@ -59,24 +53,15 @@ impl ApiWorker {
...
-
-#[async_trait]
-impl Actor for ApiWorker {
-    type State = ();
-
-    type Error = ApiError;
 
-    async fn init(&mut self, cx: &mut ActorContext<Self>) -> Result<Self::State, Self::Error> {
-        let (sender, receiver) = oneshot::channel();
+    // TODO: What happens when there is an error in an API request. This should not bring down all of Chronicle.
```
Alexandcoats, line 59: I don't think errors in requests will ever stop the axum server.

grtlr: Ok, great!

src/bin/inx-chronicle/api/mod.rs
```diff
@@ -57,7 +58,10 @@ impl ApiWorker {
     }
 
     // TODO: What happens when there is an error in an API request. This should not bring down all of Chronicle.
-    pub async fn run(&self) -> Result<(), ApiError> {
+    pub async fn run<Fut>(&self, shutdown_handle: Fut) -> Result<(), ApiError>
```
grtlr, line 61: Maybe use `impl Future` and skip the `.then` part below?

Alex6323:
Bring back the environment stuff for specifying the location of the config.

src/bin/inx-chronicle/error.rs
```diff
@@ -17,4 +17,6 @@ pub enum Error {
     #[cfg(feature = "inx")]
     #[error(transparent)]
     Inx(#[from] super::stardust_inx::InxError),
+    #[error(transparent)]
+    Shutdown(#[from] tokio::sync::broadcast::error::SendError<()>),
```
grtlr, line 21: Maybe we can just use a panic, instead of introducing a new error variant?

src/bin/inx-chronicle/api/mod.rs
Alexandcoats: The `.then` is because the return type is specified in the graceful shutdown fn, so it's not avoidable. But we can use `impl Future`.

src/bin/inx-chronicle/error.rs
Alexandcoats: Let's just assume we're going to refactor later using a different error lib.

Alexandcoats (approved):

Alex6323 (dismissed):

Alex6323 (approved):

PR #779, opened 2022-10-06
refactor(db): make `MongoDbCollection` trait async and move `create_indexes` in there

Alex6323:
## PR Description
Currently `create_indexes` is implemented conventionally for each concrete collection type that requires indexes for performance reasons, e.g. `BlockCollection`, `OutputCollection`, etc. Even for those that are currently not supported by indexes this may change in the future and such a method would then have to be added. This PR makes `create_indexes` part of our `MongoDbCollection` trait with a default empty body in case a particular collection doesn't need additional indexes. Since this method is async this also implies making `MongoDbCollection` an async trait.

## Motivation

(a) it makes sense from a design perspective
(b) it allows writing generic helper functions in our test code

## Notes to Reviewer

#### Core Changes
* [ ] Check that all indexes are still properly used.
* [ ] Run chronicle.
* [ ] Run tests.


Alexandcoats:
Can you move the create collection stuff from the launcher to the `create_index` methods?

grtlr (approved):
Looks good. If #699 depends on it you can rebase it onto that one, than we can quickly merge them one after another (still waiting for the review of @Alexandcoats).

src/db/mongodb/mod.rs
```diff
@@ -56,8 +56,10 @@ impl MongoDb {
     }
 
     /// Creates a collection if it does not exist.
-    pub async fn create_collection<T: MongoDbCollection>(&self) {
+    pub async fn create_indexes<T: MongoDbCollection + Send + Sync>(&self) -> Result<(), Error> {
```
grtlr, line 59: At this point `initialize` might be a better name?

Alex6323: yea, might be ... `initialize` is a bit opaque though and according to the docs `self.db.create_collection` how we call it (without any options) is pretty much a no-op (as MongoDb creates collections lazily when you start inserting). I think @Alexandcoats still thinks it's a necessary call so I kept it, but the true purpose of this call is in fact to create the indexes.

Alex6323: However ... i don't care so much, I just wanted to keep the current "naming convention".

grtlr: It's important to call `create_collection` because otherwise the indexes are not created at start up. I also don't mind the current name, was just a suggestion.

grtlr (dismissed):

src/db/mongodb/mod.rs
Alex6323: ah, I thought that might have been the reason ... so in that case we might want to still be able to call it (`self.db.create_collection`) separate from creating the associated indexes :thinking: and controlling this via a boolean argument.

Alexandcoats: This is not what I had in mind when I suggested moving this functionality here. I meant that the create_indexes fn in the trait should also create the collection for simplicity.

Alex6323: Not sure I understand what you want. The `create_indexes` fn in the trait is supposed to be overwritten. :thinking:

Alex6323: the collection is already created in `crate_indexes` of the `MongoDb` type before it calls the trait method. So isn't that simple enough? @Alexandcoats

Alexandcoats: I'm saying we shouldn't have `MongoDb::create_indexes`, we should have `MongoDb::create_collection` which is called by the trait's `create_indexes`.

Alex6323: alright, we can do that.

src/db/collections/outputs/mod.rs
```diff
@@ -48,24 +48,64 @@ pub struct OutputDocument {
 
 /// The stardust outputs collection.
 pub struct OutputCollection {
-    db: mongodb::Database,
+    db: MongoDb,
```
Alexandcoats, line 51: Why make this change?

Alex6323: I only made this change due to what you said here:
https://github.com/iotaledger/inx-chronicle/pull/779#discussion_r990251965
:shrug:

grtlr (approved):

PR #780, opened 2022-10-06
feat(db): add query profiler

### LINKED ISSUE #713: Emit warnings for slow queries
We should investigate if we can display warnings from MongoDB's slow query log, similar to how Hornet is doing this for the Indexer.

Alex6323:
Things I still want to do:
* [ ] test it with various values
* [ ] enable/disable and configure it in the config 
* [ ] investigate grafana/prometheus support
 
## Linked Issues

* Closes #713


grtlr:
I'm closing this for now as there is no real way forward with this.

Issue #704, opened 2022-09-20
Find optimizations for and cleanup stream iterators


grtlr:
I think we should be able to gain performance in syncing by performing concurrent writes for chunks. Currently, we are only doing this for the initial ledger state.

Since we use this in multiple places we should come up with a good abstraction that can be reused.

Alexandcoats:
Look into reusing code for stream compositions in the INX worker.

Alex6323:
I am still looking into it. The issue is not specific enough IMO to be actionable.

Alexandcoats:
We already have about as good performance from the streams as we'll get now that they all operate in parallel. The only thing potentially remaining is to consolidate code, but tbh I don't think it's possible to do it in a clean way. Or, rather, the current method using stream mappers is the cleanest (and most idiomatic) way.

grtlr:
Yeah, I don't really see a way forward (nor a necessity). Closing this for now.

PR #782, opened 2022-10-06
feat(inx): remove `LedgerUpdateStream` and manual chunks iter

### LINKED ISSUE #689: Remove the `ChunksExt` iterator and use `itertools::Chunks`
With the actor system removed, we probably don't need the `ChunksExt` iterator anymore and can use `itertools::Chunks` instead.

### LINKED ISSUE #702: Remove `LedgerUpdateStream`

Alexandcoats:
## Linked Issues

* Closes #702
* Closes #689 

## Notes to Reviewer

#### INX Changes
* [x] Measure performance differences between previous and new implementation.


grtlr:
Any reason why you choose not to do the handling similar to: https://github.com/grtlr/nort/blob/main/src/inx.rs#L22-L57? I think doing things this way would allow us to write a generic, iterator-based, chunk inserter that would be easier to test and reuse.

grtlr:
Here we should assume it though because it's guaranteed to be like that by INX: https://github.com/iotaledger/inx/blob/develop/proto/inx.proto#L37-L41

Alexandcoats:
> Here we should assume it though because it's guaranteed to be like that by INX: https://github.com/iotaledger/inx/blob/develop/proto/inx.proto#L37-L41

Sure, but it's not guaranteed by the API. Since it's not an issue to handle it this way, I don't think we should change it.

> I'd much rather use the regular `chunks` method instead of rolling our own.

As I have stated in that issue, it is not possible.

> Also, regarding the iterator implementation, I was refering to this: https://github.com/iotaledger/inx-chronicle/issues/704

We could make something generic, but why? This is used in one place. Also, my impl parallelizes the places that can be.

src/bin/inx-chronicle/stardust_inx/mod.rs
```diff
@@ -75,12 +72,12 @@ impl InxWorker {
...
         let (start_index, mut inx) = self.init().await?;
 
-        let mut stream = LedgerUpdateStream::new(inx.listen_to_ledger_updates((start_index.0..).into()).await?);
+        let mut stream = inx.listen_to_ledger_updates((start_index.0..).into()).await?;
 
         debug!("Started listening to ledger updates via INX.");
 
         while let Some(ledger_update) = stream.try_next().await? {
-            handle_ledger_update(ledger_update, &mut inx, &self.db).await?;
+            handle_ledger_update(&mut inx, &self.db, ledger_update, &mut stream).await?;
```
Alex6323, line 80: at this point we can just move the `while let` into `handle_ledger_update` IMO

src/bin/inx-chronicle/stardust_inx/mod.rs
```diff
@@ -213,46 +211,109 @@ impl InxWorker {
...
+        .try_chunks(INSERT_BATCH_SIZE)
+        // We only care if we had an error, so discard the other data
+        .map_err(|e| e.1)
+        // Convert batches to tasks
+        .try_fold(&mut tasks, |tasks, batch| async {
+            let db = db.clone();
+            tasks.spawn(async move { insert_unspent_outputs(&db, &batch).await });
+            Result::<_, InxError>::Ok(tasks)
+        })
+        .await?;
```
Alex6323, line 279: make this DRY with a macro?

src/bin/inx-chronicle/stardust_inx/mod.rs
Alexandcoats: Well the reason I did it this way is to keep the span around the function.

src/bin/inx-chronicle/stardust_inx/mod.rs
Alexandcoats: Oh hell no

Alex6323: :joy:

src/bin/inx-chronicle/stardust_inx/mod.rs
Alex6323: oh I see ... damn it!

Alex6323 (dismissed):

PR #811, opened 2022-10-17
feat(error): clean up errors with `eyre` lib

### LINKED ISSUE #844: Investigate better error management

Alexandcoats:
Closes #844.

## Notes to Reviewer

#### Config Changes
* [x] Ensure that config changes work with individual build features.

#### API Changes

#### INX Changes


Alex6323:
tbh...on first review...I am a bit disappointed in what using `eyre` gives us. Agreed ... the error enums become less verbose which is nice, but it seems like we are buying this with lots' `map_err`s, and hence more lines of code in the logic we actually have to look at more often. In a cleanup PR I would expect a net-minus of LOC, but instead LOC increased. Are there other benefits I am overlooking atm?

grtlr (changes requested):
For me this crashes with the following error:

```
ERROR inx_chronicle: panicked at 'a handler must always be installed if the `auto-install` feature is disabled', src/bin/inx-chronicle/main.rs:53:9
```

Any idea on what could be the problem here?

src/bin/inx-chronicle/api/error.rs
```diff
@@ -3,141 +3,187 @@
...
-    MongoDb(#[from] mongodb::error::Error),
-    #[error(transparent)]
-    PasswordHash(#[from] auth_helper::password::Error),
-    #[error(transparent)]
-    UrlEncoding(#[from] serde_urlencoded::de::Error),
-    #[error(transparent)]
-    ValueAccess(#[from] ValueAccessError),
+pub enum CorruptStateError {
+    #[error("no milestone in the database")]
+    Milestone,
```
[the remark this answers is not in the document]
Alexandcoats: This would go against the format of the rest of the variants.

grtlr (changes requested):
We should return error code `4**` for data that is not found in the database, the current internal errors (code `500`) needlessly pollute our logging output.

src/bin/inx-chronicle/api/error.rs
```diff
@@ -4,140 +4,205 @@
...
-impl<T: Into<InternalApiError>> From<T> for ApiError {
-    fn from(err: T) -> Self {
-        ApiError::Internal(err.into())
+#[derive(Error, Debug)]
+#[allow(missing_docs)]
+pub enum MissingError {
+    #[error("no results returned")]
+    NoResults,
+    #[error("no endpoint found")]
+    NotFound,
```
Alex6323, line 144: I personally think that `NotFound` is not clear enough IMO (you could also say that about data). How about "NoEndpoint", or `EndpointNotFound` to make it unambiguous?

src/bin/inx-chronicle/api/error.rs
```diff
@@ -4,140 +4,205 @@
...
-    #[error(transparent)]
+    #[error("invalid secret key: {0}")]
     SecretKey(#[from] super::secret_key::SecretKeyError),
-    #[error(transparent)]
-    TimeConversion(#[from] time::error::ConversionRange),
 }
+
+impl ErrorStatus for ConfigError {
+    fn status(&self) -> StatusCode {
+        StatusCode::BAD_REQUEST
```
Alex6323, line 202: Are we sure about this? Shouldn't that be a `5**`? :thinking:

src/bin/inx-chronicle/api/error.rs
Alexandcoats: Talk to the HTTP protocol creators?

src/bin/inx-chronicle/api/error.rs
Alexandcoats: Hmm...perhaps.

src/bin/inx-chronicle/api/error.rs
Alex6323: Well ... the rebellion has to start somewhere :trollface:

grtlr (approved):

Alex6323 (approved):

PR #816, opened 2022-10-20
feat(types): improve and test encoding of `KIND`

grtlr:
## Linked Issues

None

## Notes to Reviewer

#### Test cases

### DIFF
--- a/src/db/collections/mod.rs
+++ b/src/db/collections/mod.rs
@@ -45,7 +45,7 @@ pub trait OutputKindQuery {
 
 impl OutputKindQuery for () {}
 
-macro_rules! impl_output_kind {
+macro_rules! impl_output_kind_query {
     ($t:ty) => {
         impl OutputKindQuery for $t {
             fn kind() -> Option<&'static str> {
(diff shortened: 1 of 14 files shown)


src/db/collections/mod.rs
```diff
@@ -32,56 +32,34 @@ pub use self::{
...
 };
 
 /// Helper to specify a kind for an output type.
-pub trait OutputKind {
+pub trait OutputKindQuery {
     /// Gets the output kind.
-    fn kind() -> Option<&'static str> {
-        None
-    }
+    fn kind() -> Option<&'static str>;
```
Alex6323, line 41: why not simply make this a `const KIND: Option<&'static str>;` ?

Alex6323, line 41: what about making this:
```rust
const KIND: &'static str;
```
and let `Output` KIND be `""`?

grtlr: No that won't work with the query building.

Alexandcoats (approved):
much improvement, very wow

PR #830, opened 2022-10-27
feat(cli)!: group CLI arguments and change INX config

### LINKED ISSUE #727: More flexible syncing by providing administrative functionality
Right now, we only offer the option to specify a syncing milestone from which INX starts syncing. 

We should create a new `[sync]` section in the `Config` that gives users more flexibility:

* `sync_start_milestone = 1` should move to that section
* Add an additional option `delete_newer` that would delete any data that is newer than that milestone

The motivation behind this is, that if there was a problem with Chronicle and we know a safe milestone (like 1 week ago, the milestone before the update) the user can only do a partial wipe of the DB and does not have to resync everything from scratch.

Alex6323:
## Linked Issues

* Closes #727 

Restructures our CLI into argument groups (not `ArgGroup`s though) which keeps them logically separated and easier to feature gate.

## Notes to Reviewer

#### Config Changes
* [x] Ensure that config changes work with individual build features.


src/bin/inx-chronicle/cli.rs
```diff
@@ -15,35 +15,39 @@ pub struct ClArgs {
     /// The location of the configuration file.
     #[arg(short, long, env = "CONFIG_PATH")]
     pub config: Option<String>,
-    /// The MongoDB connection string.
-    #[arg(long, env = "MONGODB_CONN_STR")]
-    pub mongodb_conn_str: Option<String>,
     /// The url pointing to an InfluxDb instance.
-    #[arg(long, env = "INFLUXDB_URL")]
+    #[arg(long = "influxdb.url", env = "INFLUXDB_URL")]
```
Alexandcoats, line 19: Nope, don't like this.

Alex6323: you mean the dot notation?

Alexandcoats: Yeah but they probably have a reason for that, I.E. they parse the dot notation to set variables directly in the config or something. We don't.

Alex6323: Changed

src/bin/inx-chronicle/cli.rs
```diff
@@ -15,39 +15,90 @@ pub struct ClArgs {
...
-    #[command(subcommand)]
-    pub subcommand: Option<Subcommands>,
+}
+
+#[cfg(feature = "inx")]
+#[derive(Args, Debug)]
+#[command(group = ArgGroup::new("inx"))]
+pub struct InxArgs {
+    /// Toggle INX write workflow.
+    #[arg(long = "inx-enabled", env = "INX_ENABLED", group = "inx")]
```
Alexandcoats, line 69: Any reason why we need to specify the `long`s? They are automatically generated.

Alex6323: remainder of the former "dot attempt". Sometimes it's necessary though as for the interface (aimed at the user) there might be something else more appropriate than just the converted field name. I'll change the places where it's the same.

src/bin/inx-chronicle/cli.rs
```diff
@@ -15,39 +15,90 @@ pub struct ClArgs {
...
+    /// The address of the INX interface provided by the node.
+    #[arg(long, env = "INX_URL")]
+    pub inx_url: Option<String>,
+    /// Milestone at which synchronization should begin. A value of `1` means syncing back until genesis (default).
+    #[arg(long = "inx-sync-start")]
+    pub sync_start: Option<u32>,
+}
+
+#[derive(Args, Debug)]
+#[command(group = ArgGroup::new("mongodb").args(["mongodb_conn_str"]))]
```
Alexandcoats, line 80: What is the purpose of the arg groups?

grtlr (dismissed):

grtlr (approved):

Alexandcoats (approved):

PR #854, opened 2022-11-07
feat(api): add proof-of-inclusion (PoI) endpoints

### LINKED ISSUE #849: Tracking Issue - `Proof-of-Inclusion`
### About tracking issues

Tracking issues are used to record the overall progress of implementation.
They are also uses as hubs connecting to other relevant issues, e.g., bugs or open design questions.
A tracking issue is however *not* meant for large scale discussion, questions, or bug reports about a feature.
Instead, open a dedicated issue for the specific matter.

### Steps

* [x] #850
* [x] #851
* [x] #852
* [x] #890

### Unresolved Questions

There might be a better way to store a proofs than the current format.

### Implementation history

* Here is a proof of concept implementation: https://github.com/iotaledger/inx-poi

Alex6323:
## Linked Issues

* Closes #849 

## Notes to Reviewer

#### API Changes

#### Test cases
* [x] Review and run tests that were added/changed.
* [x] Suggest places that may benefit from test cases.

#### INX Changes

#### Database Changes

#### Additionally
* [x] Run the `inx-poi` Hornet plugin and compare the generated proofs of inclusion. It should be possible to validate a PoI with Chronicle that was generated by `inx-poi` and vice versa.


documentation/api/api-poi.yaml
```diff
@@ -0,0 +1,284 @@
...
+  url: https://iota.org
+servers:
+  - url: http://127.0.0.1:14265
+tags:
+  - name: create
+    description: Proof-of-Inclusion creation.
+  - name: validate
+    description: Proof-of-Inclusion validation.
+paths:
+  /api/poi/v1/create/{blockId}:
```
Alexandcoats, line 22: Perhaps it would be better to use `/api/proof/...`?

grtlr: I think he's following the definition of `inx-poi` here.

Alex6323: yeah, that's correct and it would be confusing for people if we use a different routes than they do.

src/bin/inx-chronicle/api/stardust/poi/error.rs
```diff
@@ -4,12 +4,12 @@
 use thiserror::Error;
 
 #[derive(Error, Debug)]
-#[allow(missing_docs)]
+#[allow(missing_docs, clippy::enum_variant_names)]
```
Alexandcoats, line 8: Just name them better 😆

Alex6323: Any suggestions? Because I like them how they are named. It's also a stupid clippy lint. What is it helping with exactly?

src/bin/inx-chronicle/api/stardust/poi/routes.rs
```diff
@@ -159,7 +160,10 @@ fn get_valid_public_keys_for_index(mut key_ranges: Box<[MilestoneKeyRange]>, ind
         match (key_range.start, key_range.end) {
             (start, _) if start > index => break,
             (start, end) if index <= end || start == end => {
-                public_keys.insert(key_range.public_key.clone());
+                // Panic: should never fail
```
Alexandcoats, line 163: 🤔

src/bin/inx-chronicle/api/stardust/poi/error.rs
Alexandcoats: Just remove the prefix `Invalid`.

Alex6323: I know that this is what I am supposed to do, but what if I don't agree? It would be inconsistent with how we name other error variants: for instance `InxError`. Should we remove `Invalid` and `Missing` there, too? This clippy warning should belong to the pedantic group of suggestions as it doesn't help with improving the code in any significant way.

I am fixing this in another way.

src/bin/inx-chronicle/api/stardust/poi/routes.rs
Alex6323: I looked up the error conditions...if we trust Hornet and our own database then this won't fail.

Cargo.toml
```diff
@@ -133,6 +136,10 @@ metrics = [
     "influxdb",
     "dep:chrono",
 ]
+poi = [
```
Alexandcoats, line 139: Should this be default?

Alexandcoats (changes requested):
When I send an invalid proof, I get an error rather than `valid: false`. For example, changing the milestone index or a block ID randomly.

src/db/collections/block.rs
```diff
@@ -188,6 +189,30 @@ impl BlockCollection {
...
+    pub async fn get_pastcone_in_white_flag_order(&self, index: MilestoneIndex) -> Result<Vec<BlockId>, Error> {
+        #[derive(Deserialize)]
+        struct BlockIdResult {
+            id: BlockId,
+        }
+
+        let block_ids = self
+            .aggregate::<BlockIdResult>(
+                vec![
+                    doc! { "$match": { "metadata.referenced_by_milestone_index": index } },
```
Alexandcoats, line 202: Shouldn't we provide the white flag of the requested block and only get blocks with `"white_flag_index": { "$lt": index }`?

src/bin/inx-chronicle/api/stardust/poi/merkle_proof.rs
```diff
@@ -0,0 +1,248 @@
...
+    pub fn create_proof(&self, block_ids: &[BlockId], block_id: &BlockId) -> Result<MerkleProof, PoIError> {
+        let index = find_index(block_ids, block_id).ok_or(PoIError::InvalidRequest("invalid BlockId"))?;
+        self.create_proof_from_index(block_ids, index)
+    }
+
+    // NOTE: `block_ids` is the list of past-cone block ids in "White Flag" order.
+    fn create_proof_from_index(&self, block_ids: &[BlockId], index: usize) -> Result<MerkleProof, PoIError> {
+        let n = block_ids.len();
+        if n < 2 {
+            Err(PoIError::InvalidInput("cannot create proof for less than 2 block ids"))
```
Alexandcoats, line 66: These should be error variants.

src/bin/inx-chronicle/api/stardust/poi/routes.rs
Alexandcoats: Still, returning an error isn't difficult and it would print a better error just in case.

src/bin/inx-chronicle/api/stardust/poi/merkle_proof.rs
```diff
@@ -0,0 +1,248 @@
...
+                    }
+                };
+                let hash = proof.hash(self);
+                Hashable::MerkleProof(Box::new(proof), hash)
+            }
+        }
+    }
+}
+
+fn find_index(block_ids: &[BlockId], block_id: &BlockId) -> Option<usize> {
```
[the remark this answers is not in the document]
Alex6323: yeah, sometimes I like to do this for documentation purposes to create some prosa without comments. I am very certain that this gets automatically inlined by the compiler even without adding an attribute.

src/bin/inx-chronicle/api/stardust/poi/routes.rs
Alex6323: But think again about what this means! We just pulled those coordinator public keys from our database and they cannot be decoded. That's not something that is going to change with another API call. If that `unwrap` panics Chronicle's db got corrupted.

src/db/collections/block.rs
Alex6323: That would only be possible if we would have pre-computed and stored all the hashes of the merkle tree. Currently we need to do that on request and I think that's probably the better approach.

Cargo.toml
Alex6323: I think there's no real reason why it couldn't tbh. Wdyt @grtlr ?

src/bin/inx-chronicle/api/stardust/poi/routes.rs
Alexandcoats: We won't shutdown either way, this is the API.

src/db/collections/block.rs
```diff
@@ -188,6 +189,30 @@ impl BlockCollection {
...
+                    doc! { "$sort": { "metadata.white_flag_index": 1 } },
```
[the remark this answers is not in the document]
Alexandcoats: Maybe we should add it as part of the `referenced_by_milestone_index` index.

Alexandcoats:
Two differences I noticed when comparing `inx-poi` and our impl:
1. We include `metadata` and `stateMetadata` that does not exist in their response when it is `0x`.
2. We hash the `value` and they do not (it is just the requested block ID).

I think the 1st is fine, but we probably should change the 2nd.

Cargo.toml
grtlr: Yeah, let's make it default 👍

src/db/collections/block.rs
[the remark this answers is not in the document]
grtlr: That's a good idea!

Cargo.toml
Alex6323: Done

src/bin/inx-chronicle/api/stardust/poi/merkle_proof.rs
Alex6323: Correct

src/bin/inx-chronicle/api/stardust/poi/routes.rs
Alex6323: that's a good point ... it should be a big red error in the logs but no shutdown.

src/bin/inx-chronicle/api/stardust/mod.rs
```diff
@@ -6,10 +6,21 @@ use super::router::Router;
...
-        .nest("/indexer/v1", indexer::routes())
+        .nest("/explorer/v2", explorer::routes())
+        .nest("/indexer/v1", indexer::routes());
+
+    #[cfg(feature = "poi")]
+    {
+        router = router.nest("/poi/v1", poi::routes());
+    }
+
+    router
```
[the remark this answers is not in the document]
Alexandcoats: Well, there's a reason we don't use it like this. Because the macro doesn't remove branches at compile time, this will fail to compile without the `poi` feature because the `poi` module won't exist.

src/bin/inx-chronicle/api/stardust/poi/merkle_proof.rs
```diff
@@ -81,10 +85,23 @@ impl MerkleProof {
         self.compute_audit_path(&data, index)
     }
 
-    /// Calculates the "Merkle Audit Path"
+    /// Recursively computes the "Merkle Audit Path" for a certain `BlockId` that is given by its index in a list of
+    /// ordered and unique `BlockId`s.
+    ///
+    /// Given `n` block ids, the algorithm deterministically selects a `mid` (largest power of two
```
Alexandcoats, line 91: When I suggested adding this kind of detailed comment, I didn't mean in the docs...This is for us, which is why I just used regular comments.

src/bin/inx-chronicle/api/stardust/mod.rs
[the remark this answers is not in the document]
Alex6323: exactly....I wanted to use the `cfg!` macro several times now (also in Bee) only to find that you can make use of it relatively rarely.

src/db/collections/block.rs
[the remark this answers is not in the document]
grtlr: @Alex6323 has this been resolved already?

src/bin/inx-chronicle/api/stardust/poi/merkle_proof.rs
Alexandcoats: I really cannot understand your reasoning here.

Alex6323: that's unfortunate :see_no_evil: 

Then I try again to make my point... I don't think that a single line of this function needs explanation. What you really want is an explanation of how the overall algorithm works...at least that's what I understood from our discussions. Hence, IMHO the explanation should be outside the method body be it as a doc comment or a normal comment.

src/db/collections/block.rs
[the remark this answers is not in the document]
Alex6323: yup, already resolved ... I added a compound index.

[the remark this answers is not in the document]
Alexandcoats: Uhhhh...where?

src/bin/inx-chronicle/api/stardust/poi/routes.rs
```diff
@@ -91,11 +97,15 @@ async fn create_block_referenced_proof(database: Extension<MongoDb>, Path(block_
         .into());
     }
 
-    Ok(CreateProofResponse {
-        milestone: milestone.into(),
-        block: block.into(),
-        audit_path: merkle_audit_path.into(),
-    })
+    if matches!(headers.get(ACCEPT), Some(header) if header == BYTE_CONTENT_HEADER) {
+        todo!("return proof in binary format");
```
Alexandcoats, line 101: Just a reminder comment

Alex6323: thanks, but after some consideration the idea of returning a binary response has been dropped for now until there's a use-case for it. The JSON response can easily be converted to whatever the user wants/requires client-side. If there's need for compression then this should be done client-side IMO.

grtlr (approved):

PR #866, opened 2022-11-07
feat(cli): add helpful CLI commands

### LINKED ISSUE #845: Add a CLI command to clear the database
## Description

There should be a debug command to clear the database, available only using debug builds. This can be done using:

```rust
#[cfg(debug_assertions)]
```

## Requirements

The command should probably just connect to MongoDb using the same process as the binary and drop the database.

Yes.

Alexandcoats:
## Linked Issues

* Closes #845 

## Notes to Reviewer

#### CLI Changes
- [x] Run the new CLI commands and verify that they are successful

### DIFF
--- a/src/bin/inx-chronicle/cli.rs
+++ b/src/bin/inx-chronicle/cli.rs
@@ -152,7 +152,7 @@ impl ClArgs {
     /// Process subcommands and return whether the app should early exit.
     #[allow(unused)]
     #[allow(clippy::collapsible_match)]
-    pub async fn process_subcommands(&self, config: &ChronicleConfig) -> Result<bool, Error> {
+    pub async fn process_subcommands(&self, config: &ChronicleConfig) -> Result<PostCommand, Error> {
         if let Some(subcommand) = &self.subcommand {
             match subcommand {
                 #[cfg(feature = "api")]
@@ -176,7 +176,7 @@ impl ClArgs {
                         exp_ts,
                         humantime::format_duration(api_data.jwt_expiration)
                     );
-                    return Ok(true);
+                    return Ok(PostCommand::Exit);
                 }
                 #[cfg(feature = "analytics")]
                 Subcommands::FillAnalytics {
@@ -184,6 +184,7 @@ impl ClArgs {
                     end_milestone,
                     num_tasks,
                 } => {
+                    println!("Connecting to database at bind address `{}`.", config.mongodb.conn_str);
                     let db = chronicle::db::MongoDb::connect(&config.mongodb).await?;
                     let start_milestone = if let Some(index) = start_milestone {
                         *index
@@ -231,12 +232,29 @@ impl ClArgs {
                         // Panic: Acceptable risk
                         res.unwrap()?;
                     }
-                    return Ok(true);
+                    return Ok(PostCommand::Exit);
+                }
+                #[cfg(debug_assertions)]
+                Subcommands::ClearDatabase { no_run } => {
+                    println!("Connecting to database at bind address `{}`.", config.mongodb.conn_str);
+                    let db = chronicle::db::MongoDb::connect(&config.mongodb).await?;
+                    db.clear().await?;
+                    println!("Database cleared successfully.");
+                    if *no_run {
+                        return Ok(PostCommand::Exit);
+                    }
+                }
+                Subcommands::BuildIndexes => {
+                    println!("Connecting to database at bind address `{}`.", config.mongodb.conn_str);
+                    let db = chronicle::db::MongoDb::connect(&config.mongodb).await?;
+                    super::build_indexes(&db).await?;
+                    println!("Indexes built successfully.");
+                    return Ok(PostCommand::Exit);
                 }
                 _ => (),
             }
         }
-        Ok(false)
+        Ok(PostCommand::Start)
     }
 }
 
@@ -254,4 +272,16 @@ pub enum Subcommands {
         #[arg(short, long)]
         num_tasks: Option<usize>,
     },
+    #[cfg(debug_assertions)]
+    ClearDatabase {
+        #[arg(short, long)]
+        no_run: bool,
+    },
+    BuildIndexes,
+}
+
+#[derive(Copy, Clone, PartialEq, Eq)]
+pub enum PostCommand {
+    Start,
+    Exit,
 }
--- a/src/bin/inx-chronicle/cli.rs
+++ b/src/bin/inx-chronicle/cli.rs
@@ -235,12 +235,12 @@ impl ClArgs {
                     return Ok(PostCommand::Exit);
                 }
                 #[cfg(debug_assertions)]
-                Subcommands::ClearDatabase { no_run } => {
+                Subcommands::ClearDatabase { run } => {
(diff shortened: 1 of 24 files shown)


src/bin/inx-chronicle/cli.rs
```diff
@@ -254,4 +272,16 @@ pub enum Subcommands {
...
+        no_run: bool,
+    },
+    BuildIndexes,
+}
+
+#[derive(Copy, Clone, PartialEq, Eq)]
+pub enum PostCommand {
+    Start,
+    Exit,
 }
```
grtlr, line 287: [nitpick] Should we use [`std::ops::ControlFlow`](https://doc.rust-lang.org/std/ops/enum.ControlFlow.html) here?

src/bin/inx-chronicle/cli.rs
```diff
@@ -231,12 +232,29 @@ impl ClArgs {
...
+                    println!("Database cleared successfully.");
+                    if *no_run {
+                        return Ok(PostCommand::Exit);
+                    }
+                }
+                Subcommands::BuildIndexes => {
+                    println!("Connecting to database at bind address `{}`.", config.mongodb.conn_str);
+                    let db = chronicle::db::MongoDb::connect(&config.mongodb).await?;
+                    super::build_indexes(&db).await?;
+                    println!("Indexes built successfully.");
```
grtlr, line 251: It's kind of weird to have a mixture between `println!` and `tracing` statements (`info!` within `super::build_indexes` for example). I agree that CLI commands lend themselves more to `println!` statements, but I don't see any way to support this in all called functions without introducing a lot of overhead. Should we switch to `tracing::info!` here instead then?

src/bin/inx-chronicle/cli.rs
Alexandcoats: Eh? I don't really care, but that enum has more than we need for this.

src/bin/inx-chronicle/cli.rs
Alexandcoats: Yeah, probably.

src/bin/inx-chronicle/cli.rs
grtlr: At some point we might want to transport information to the main process though. We can just set `B` and `C` to `()` for now.

grtlr (dismissed):

src/bin/inx-chronicle/cli.rs
```diff
@@ -170,20 +170,21 @@ impl ClArgs {
...
-                    return Ok(true);
+                    return Ok(PostCommand::Exit);
                 }
                 #[cfg(feature = "analytics")]
                 Subcommands::FillAnalytics {
                     start_milestone,
                     end_milestone,
                     num_tasks,
                 } => {
+                    tracing::info!("Connecting to database at bind address `{}`.", config.mongodb.conn_str);
```
Alex6323, line 187: tbh .. that's not a good idea as the `conn_str` can contain the credentials. You should try to extract the `url` from it.

src/bin/inx-chronicle/cli.rs
```diff
@@ -219,9 +220,9 @@ impl ClArgs {
                                 {
                                     let analytics = db.get_all_analytics(index).await?;
                                     influx_db.insert_all_analytics(timestamp, index, analytics).await?;
-                                    println!("Finished analytics for milestone: {}", index);
+                                    tracing::info!("Finished analytics for milestone: {}", index);
                                 } else {
-                                    println!("No milestone in database for index {}", index);
+                                    tracing::info!("No milestone in database for index {}", index);
```
Alex6323, line 225: one contains a `:`, the other doesn't

src/bin/inx-chronicle/cli.rs
```diff
@@ -231,12 +232,29 @@ impl ClArgs {
                         // Panic: Acceptable risk
                         res.unwrap()?;
                     }
-                    return Ok(true);
+                    return Ok(PostCommand::Exit);
+                }
+                #[cfg(debug_assertions)]
+                Subcommands::ClearDatabase { no_run } => {
+                    tracing::info!("Connecting to database at bind address `{}`.", config.mongodb.conn_str);
```
Alex6323, line 239: see comment above

src/bin/inx-chronicle/cli.rs
```diff
@@ -231,12 +232,29 @@ impl ClArgs {
...
+                    let db = chronicle::db::MongoDb::connect(&config.mongodb).await?;
+                    db.clear().await?;
+                    tracing::info!("Database cleared successfully.");
+                    if *no_run {
+                        return Ok(PostCommand::Exit);
+                    }
+                }
+                Subcommands::BuildIndexes => {
+                    tracing::info!("Connecting to database at bind address `{}`.", config.mongodb.conn_str);
```
Alex6323, line 248: see comment above

src/bin/inx-chronicle/cli.rs
```diff
@@ -231,12 +232,29 @@ impl ClArgs {
...
+                }
+                #[cfg(debug_assertions)]
+                Subcommands::ClearDatabase { no_run } => {
+                    tracing::info!("Connecting to database at bind address `{}`.", config.mongodb.conn_str);
+                    let db = chronicle::db::MongoDb::connect(&config.mongodb).await?;
+                    db.clear().await?;
+                    tracing::info!("Database cleared successfully.");
+                    if *no_run {
+                        return Ok(PostCommand::Exit);
+                    }
```
Alex6323, line 245: Instead of a `--no-run` flag I would do the opposite and add a `--run` flag, that says "please clear the database and also start Chronicle for me". That's because making it "opt-in" instead of "opt-out" is probably more what a user would expect from the name of that command.

src/bin/inx-chronicle/cli.rs
Alexandcoats: It doesn't contain the credentials. Those are separate config variables.

src/bin/inx-chronicle/cli.rs
Alexandcoats: Well, keep in mind this is really just for us. And in my testing I have found it is most useful to run it when starting the application during debugging.

src/bin/inx-chronicle/cli.rs
grtlr: We could parse it into a MongoDB connection String (https://docs.rs/mongodb/2.3.1/mongodb/options/struct.ConnectionString.html) and extract the URL from there. Maybe we create a method for it on `MongoDbConfig`.

src/bin/inx-chronicle/cli.rs
grtlr: I agree with Alex here, I wouldn't expect the command to automatically run Chronicle.

Alexandcoats: > I agree with Alex here

You tryin to be funny, funny guy?

Alex6323: :popcorn:

grtlr: I agree with both of you ❤️

grtlr (approved):

PR #887, opened 2022-11-15
fix(analytics)!: correctly calculate alias and NFT analytics

### LINKED ISSUE #847: Fix performance problems with querying analytics
### Bug description

Querying for analytics is slow, especially for real-life Shimmer data. Most slow queries come from querying the ledger state at a given milestone index.

A possible fix for this would be to add a compound index on the `booked` and `spent` milestone indexes of an output's metadata.

### LINKED ISSUE #880: `governor_changed_count` analytic is incorrect
### Bug description

Currently we don't track ownership so we count outputs multiple times. Also, a lot of the statistics for outputs don't account for things that happened entirely in a milestone.

Alexandcoats:
## Linked Issues

* Closes #880
* Closes #847 

## Notes to Reviewer

#### Test cases
* [x] Review and run tests that were added/changed.

#### INX Changes

#### Database Changes


src/types/stardust/block/output/foundry.rs
```diff
@@ -21,6 +21,13 @@ impl FoundryId {
     const LENGTH: usize = iota::FoundryId::LENGTH;
 }
 
+impl FoundryId {
+    /// Get an implicit (zeroed) foundry ID, for new foundry outputs.
+    pub fn implicit() -> Self {
+        Self([0; Self::LENGTH])
+    }
+}
+
```
grtlr, line 30: I'm sorry for leading you down the wrong path, but foundries don't have an implicit ID. Only aliases and NFTs do. We therefore also need to find a different way to compute _created_ foundries.

grtlr (changes requested):

src/db/collections/outputs/mod.rs
```diff
@@ -574,106 +558,116 @@ mod analytics {
...
+                                } },
+                            ],
+                            "changed": [
+                                { "$match": { "output.alias_id": { "$ne": AliasId::implicit() } } },
+                                // Group by state indexes to find where it changed
+                                { "$group": {
+                                    "_id": { "alias_id": "$output.alias_id", "state_index": "$output.state_index" },
+                                    "total": { "$sum": { "$cond": [ { "$eq": [ "$metadata.booked.milestone_index", index ] }, 1, 0 ] } },
+                                    "unspent": { "$max": { "$cond": [ { "$eq": [ "$metadata.spent_metadata", null ] }, 1, 0 ] } },
+                                    "prev_state": { "$max": { "$cond": [ { "$lt": [ "$metadata.booked.milestone_index", index ] }, "$output.state_index", 0 ] } },
```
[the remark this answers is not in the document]
Alexandcoats: No, the whole point of this is to get the state we were in before this milestone so we know whether the first output is a state change or not.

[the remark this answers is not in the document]
grtlr: Ah, I see. Thanks!

grtlr (approved):

src/db/collections/outputs/mod.rs
```diff
@@ -502,178 +500,119 @@ mod analytics {
...
+                .aggregate(
+                    vec![
+                        doc! { "$match": {
+                            "output.kind": "nft",
+                            "$or": [
+                                { "metadata.booked.milestone_index": index },
+                                { "metadata.spent_metadata.spent.milestone_index": index },
+                            ],
+                        } },
+                        doc! { "$facet": {
```
Alex6323, line 516: no way around `$facet`? (I remember you found a solution to circumvent it last time)

Alexandcoats: The solution is writing multiple queries. However, in this case the index being used is before the facet, so performance-wise it should not be an issue.

Alex6323 (approved):

PR #899, opened 2022-11-17
feat(config)!: switch to user config override

### LINKED ISSUE #788: Chronicle should not use a hard-coded default config
### Bug description

Currently, if Chronicle is called without arguments (and with the temporarily removed feature of using an `.env` file) it will create a config with default values and start with it. This is confusing at best and most likely not what the user wants as it also contains credentials to MongoDb which in production won't be the default ones. I'ld definitely log if Chronicle is run with a default config, but probably would just abort Chronicle with an appropriate message.

Alex6323:
## Linked Issues

* Closes #788 

## Notes to Reviewer

#### Config Changes
* [ ] Ensure proper order in which CLI and config arguments are applied.
* [ ] Ensure that config changes work with individual build features by running `cargo ci-check-features`.

#### Test cases


Alexandcoats (commented):
NGL I hate this

Alex6323:
Hate-Closed :trollface:

PR #913, opened 2022-11-21
fix(db): switch from `d128` to `f64` for analytics

### LINKED ISSUE #911: Use `f64` instead of `d128` for values in analytics queries
## Description

This should simplify the queries make them faster and clean up some of the Influx-related code.

grtlr:
## Linked Issues

* Closes #911

## Notes to Reviewer

#### INX Changes

#### Database Changes

### DIFF
--- a/src/db/collections/analytics/mod.rs
+++ b/src/db/collections/analytics/mod.rs
@@ -4,7 +4,6 @@
 /// Schema implementation for InfluxDb.
 pub mod influx;
 
-use decimal::d128;
 use futures::TryFutureExt;
 use mongodb::{bson::doc, error::Error};
 use serde::{Deserialize, Serialize};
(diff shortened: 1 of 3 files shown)


shufps:
imho it's wrong to switch integer value to float :see_no_evil: This inevitably leads to incorrect results. Probably not a big problem in case of analytics.

And float always is slower than working with integers. Even if you need a "composite integer" value like 128bit that is not natively supported by the instruction set of the CPU :thinking: 

With f64 you have actually 52Bit for values and 11bit is exponent.

Not a good idea!

grtlr:
Thanks @shufps for the feedback, here are a couple of points addressing your comments:

* This PR does not change anything about the internal representation of the data. It is only concerned with what we store in InfluxDB. InfluxDB does not support `d128`, only `u64`.
* Of course, `f64` can suffer from rounding issues, but as you said this only happens for very large or very small numbers. I would argue, that for analytics this rounding is not a big problem. The rounding errors only occur for large values, in which case we are probably not interested in the exact amount of _glow_ but rather something like _M SMR_. If the frontend truncates decimals anyways, we can afford some more rounding errors.
* If we ever need precise data making the analytics functions/queries generic should not be that big of a problem.
* Finally, we would merge this PR only if it comes with significant performance improvements (operations on `f64` are faster than operations on `d128`). From my current testing it looks like the benefits are negible, so we might not even merge this.

Alexandcoats:
I do like that it makes the queries a bit cleaner.

Alex6323 (approved):

Alexandcoats:
Shouldn't we go ahead and close this?

grtlr:
Closing this for now, because the speedups were not significant.

PR #919, opened 2022-11-21
Feat/api/poi improvements

Alexandcoats:
Reimplementing those improvements from before


Alex6323:
I like certain improvements but not all of them. I will cherry-pick some of the commits if possible.

src/bin/inx-chronicle/api/stardust/poi/merkle_hasher.rs
```diff
@@ -1,54 +1,133 @@
...
-pub struct MerkleHasher<H> {
-    _phantom: PhantomData<H>,
-}
-
-impl<H: Default + Digest> MerkleHasher<H> {
-    pub fn new() -> Self {
-        Self { _phantom: PhantomData }
-    }
+/// A Merkle tree hasher.
+pub struct MerkleHasher;
```
Alex6323, line 16: I think I'll take this change (otherwise I have a hard time arguing why the hasher should remain generic :sweat_smile: ).

src/bin/inx-chronicle/api/stardust/poi/merkle_hasher.rs
```diff
@@ -1,54 +1,133 @@
...
-        let mut hasher = H::default();
+    /// Hash a subtree.
+    pub fn hash_node(l: impl AsRef<[u8]>, r: impl AsRef<[u8]>) -> Output<Blake2b256> {
+        let mut hasher = Blake2b256::default();
         hasher.update([NODE_HASH_PREFIX]);
         hasher.update(l);
         hasher.update(r);
         hasher.finalize()
     }
+
```
Alex6323, line 53: I want everything `proof` related in one file together with the corresponding types. IMO that keeps things better organized.

Alexandcoats: I think it makes more sense to define impls for types where that type is defined.

src/bin/inx-chronicle/api/stardust/poi/merkle_hasher.rs
```diff
@@ -1,54 +1,133 @@
...
+    /// Recursively compute a merkle tree.
+    fn compute_proof(data: &[[u8; BlockId::LENGTH]], index: usize) -> MerkleProof {
+        let n = data.len();
+        debug_assert!(index < n);
+        match n {
+            0 | 1 => unreachable!(),
+            // The terminating point, where we only have two values that become
+            // left and right leaves. The chosen index is a `Value` while
+            // the other is a `Node`.
```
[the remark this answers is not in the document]
Alexandcoats: This is a bit of a weird argument. It took me several hours to figure out this code well enough to write these comments. You so strongly disagree with adding helpful information that has no effect on compilation or runtime?

src/bin/inx-chronicle/api/stardust/poi/merkle_hasher.rs
```diff
@@ -1,54 +1,133 @@
...
+                // If the chosen index is in the left half of the tree,
+                // we build out that structure by calling this fn recursively.
+                // Otherwise, we simply hash the subtree and store it as a `Node`.
+                if index < k {
+                    MerkleProof {
+                        left: if data[..k].len() == 1 {
+                            Hashable::Value(Self::hash_leaf(data[0]))
+                        } else {
+                            Hashable::MerkleProof(Box::new(Self::compute_proof(&data[..k], index)))
+                        },
```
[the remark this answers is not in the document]
Alex6323: fair enough :+1:

src/bin/inx-chronicle/api/stardust/poi/merkle_hasher.rs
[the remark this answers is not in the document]
Alex6323: I am saying that describing how the algorithm works can be done in the `fn` docs as there's no single line in the algorithm that necessary to explain. It's just normal Rust, no `unsafe` code, no weird aliasing, no nothing that would need explaining. What I would do is write how the basic algorithm works in general in the `fn` docs and keep the logic itself lean.

src/bin/inx-chronicle/api/stardust/poi/merkle_hasher.rs
Alex6323: In general yes, but not here. Everything that deals with `proofs` should be in once place IMO so that for the reader everything relevant is nicely placed together. If we cannot agree on that then I'll probably create a new type `MerkleProver` or something to keep "hashing" and "proving" logically separated (proving builds on top of hashing).

Alexandcoats: In that case the hasher should be a module and the stuff you want grouped should be submodules.

src/bin/inx-chronicle/api/stardust/poi/merkle_hasher.rs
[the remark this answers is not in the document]
Alexandcoats: Yes, but what I'm saying is that it's NOT clear what it does in principal 😆 . Just because you aren't using `unsafe` does not mean it's clear. Recursive code in particular is always more difficult to reason about.

grtlr:
Many of these comments have been carried over to #854.

PR #931, opened 2022-11-24
feat(cli): add `health-check` command

Alex6323:
Just figured that could be helpful, especially since it works independently of the `api` feature being enabled.

## Notes to Reviewer

#### API Changes

#### CLI Changes
* [ ] Test `health-check` command.

#### Docker
* [ ] Test added health check command in `Dockerfile`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2231,9 +2231,9 @@ dependencies = [
 
 [[package]]
 name = "reqwest"
-version = "0.11.12"
+version = "0.11.13"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "431949c384f4e2ae07605ccaa56d1d9d2ecdb5cadd4f9577ccfab29f2e5149fc"
+checksum = "68cc60575865c7831548863cc02356512e3f1dc2f3f82cb837d7fc4cc8f3c97c"
 dependencies = [
  "base64",
  "bytes",
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -35,6 +35,8 @@ EXPOSE 8042/tcp
 # Metrics
 EXPOSE 9100/tcp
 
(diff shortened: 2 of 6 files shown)


Alexandcoats:
This is for docker, I assume? Wouldn't we define healthy for docker differently than our normal healthy?

Alex6323:
> This is for docker, I assume? Wouldn't we define healthy for docker differently than our normal healthy?

Yeah, for Docker and for a `api` feature independent way to check the health of Chronicle. I don't know whether we should have different kinds of `health` tbh :thinking: Wouldn't that be confusing. If you have a container that would be dependent on `inx-chronicle` you don't have to let it wait for it to become healthy. On the other hand if your container only wants to work with a healthy `inx-chronicle` than you have a way to make sure of it.

Alex6323:
> Iirc, Hornet also has two different definitions of healthy. With the current proposal, if Chronicle takes 3 days to fully sync, it would remain unhealthy throughout the entire time (while certainly some routes could be queried already).

As I said in my comment ... no one is forcing you to wait until the `inx-chronicle` container reports `healthy`. 

I am open for ideas around other definitions of `healthy` that are better suited here.

Alexandcoats:
I think for the purposes of docker containers, I see no value in discerning between a synced and unsynced instance. I would consider chronicle healthy as a docker instance if it is running all the configured workers. Am I missing something?

grtlr:
I would agree with @Alexandcoats here—a good metric would be if all workers started correctly. Ideally, we would take this one step further by defining a condition per worker, such as INX or DB connections estabslished. But at that point we probably have to introduce a something like a worker trait with corresponding `init` and `run` functions. I'm not sure if that's worth it for now?

Alexandcoats:
I think it's doable without that much boilerplate. At the most basic we could just have a `Arc<Mutex<HashMap<String, bool>>>` in the main thread and add each worker keyed by name when we spawn them. Then they could set the flag to true when they are ready. Just one option 🤷

grtlr:
Closing this for now.

PR #934, opened 2022-11-25
fix(grafana): use `mean` instead of `last` for times

grtlr:


Alexandcoats:
Reasoning?

grtlr:
The graph is really hard to read otherwise, especially for longer time spans. Also, `last` can obscure what actually has been happening during an interval.

Alexandcoats (approved):

Alex6323 (approved):

PR #935, opened 2022-11-28
feat(cli): allow cancellation of fill-analytics tool

Alex6323:
Since it's a potentially long-running command we probably want to be able to abort it.

### DIFF
--- a/src/bin/inx-chronicle/cli.rs
+++ b/src/bin/inx-chronicle/cli.rs
@@ -1,9 +1,13 @@
 // Copyright 2022 IOTA Stiftung
 // SPDX-License-Identifier: Apache-2.0
 
+use chronicle::db::{influxdb::InfluxDb, MongoDb};
 use clap::{Args, Parser, Subcommand};
 
-use crate::config::{ChronicleConfig, ConfigError};
+use crate::{
+    config::{ChronicleConfig, ConfigError},
+    process,
+};
 
 /// Chronicle permanode storage as an INX plugin
 #[derive(Parser, Debug)]
@@ -242,6 +246,7 @@ impl ClArgs {
                     };
                     let influx_db = chronicle::db::influxdb::InfluxDb::connect(&config.influxdb).await?;
                     let num_tasks = num_tasks.unwrap_or(1);
+                    let (shutdown_signal, _) = tokio::sync::broadcast::channel::<()>(1);
                     let mut join_set = tokio::task::JoinSet::<eyre::Result<()>>::new();
                     // Inclusive end
                     let total_milestones = end_milestone - start_milestone + 1;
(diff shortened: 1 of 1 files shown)


src/bin/inx-chronicle/cli.rs
```diff
@@ -255,55 +256,37 @@ impl ClArgs {
...
-                    while let Some(res) = join_set.join_next().await {
-                        // Panic: Acceptable risk
-                        res.unwrap()?;
+
+                    // We wait for either the interrupt signal or the completion of filling the analytics.
+                    loop {
+                        tokio::select! {
+                            _ = crate::process::interrupt_or_terminate() => {
+                                tracing::info!("received ctrl-c or terminate");
+                                shutdown_signal.send(())?;
```
grtlr, line 277: I might be a little slow today, but who is receiving this signal?

Alex6323: each created task in the `JoinSet`

grtlr: Right, thanks!

Alexandcoats, line 277: This won't await the tasks after a `ctrl-c` FYI, so they will likely still be aborted by the end of the run.

src/bin/inx-chronicle/cli.rs
```diff
@@ -255,55 +256,37 @@ impl ClArgs {
...
+                            },
+                            res = join_set.join_next() => {
```
Alexandcoats, line 279: This only awaits one task

Alex6323: you've seen the `loop`, right?

src/bin/inx-chronicle/cli.rs
Alex6323: yeah, I thought about that... but does it really matter? I can mitigate this problem by just starting at this index again, right?

src/bin/inx-chronicle/cli.rs
Alexandcoats: I did miss that, but this select should definitely not be in a loop.

src/bin/inx-chronicle/cli.rs
Alexandcoats: Well, it kind of defeats the entire purpose of this PR lol

Alex6323: Then I must have misunderstood what you meant with the term "run" (I thought you mean a run for a single milestone). The purpose of this PR is to abort all tasks spawned onto the `JoinSet` and it does just that (tested).

Alexandcoats: The outer loop prevents the issue I was describing, which I missed. However, I do not like that the select is inside that loop. Sure, the difference probably wouldn't be noticed with regular usage, but consider that this is possible:
1. 10 tasks are spawned.
2. Tasks are canceled with ctrl-c.
3. Task 1 is joined.
4. Ctrl-c is hit again before any other tasks are complete.

In this situation, having pressed `ctrl-c` twice, it should force quit. However, it is possible if pressed quickly to instead complete the `interrupt_or_terminate` future multiple times.

Alex6323: Pushed another solution that is as close at it can be to what I think you want.

src/bin/inx-chronicle/cli.rs
```diff
@@ -331,6 +299,74 @@ impl ClArgs {
...
+                        chronicle_version: std::env!("CARGO_PKG_VERSION").to_string(),
+                    })
+                    .await?;
+            }
+            tracing::info!("Finished analytics for milestone {}", index);
+        } else {
+            tracing::info!("No milestone in database for index {}", index);
+        }
+        if cancel.load(std::sync::atomic::Ordering::Relaxed) {
+            break;
```
Alexandcoats, line 364: Probably just `return` here.

src/bin/inx-chronicle/cli.rs
```diff
@@ -255,55 +257,21 @@ impl ClArgs {
...
                         start_milestone = end_milestone + 1;
                     }
-                    while let Some(res) = join_set.join_next().await {
-                        // Panic: Acceptable risk
-                        res.unwrap()?;
+
+                    tokio::select! {
+                        _ = shutdown() => {
+                            CANCEL.store(true, std::sync::atomic::Ordering::Relaxed);
+                            join_all(&mut tasks).await?
```
Alexandcoats, line 270: No need for the `join_all` fn, just join after the select.

PR #941, opened 2022-11-28
chore(deps): bump dependencies

grtlr:
Please make sure that the mapping of the key/data factors is correct. 🙏


Alexandcoats (approved):
Looks right to me, assuming everything is correct in `iota_types`.

PR #944, opened 2022-11-30
quick demo

grtlr:
just one suggestion

### DIFF
--- a/src/db/collections/outputs/mod.rs
+++ b/src/db/collections/outputs/mod.rs
@@ -718,6 +718,20 @@ mod analytics {
             prev_analytics: &mut LedgerSizeAnalytics,
             ledger_index: MilestoneIndex,
         ) -> Result<(), Error> {
+
+            let sum = || doc! { "$group" : {
+                "_id": null,
+                "total_key_bytes": { "$sum": { "$toDecimal": "$details.rent_structure.num_key_bytes" } },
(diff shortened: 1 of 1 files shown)


Alexandcoats:
What about the other function that uses identical documents?

Alexandcoats:
Yes, but then things start to really get unclean.

grtlr:
I'd argue against that. You could even write test cases for those individual functions.

Alexandcoats:
There are two problems with that:
1. We will end up with lots of free-floating functions that just generate *part* of a query and make no sense out of context. Naming them will be an issue too.
2. Separating the query into many pieces obfuscates it and makes it difficult to reason about.

grtlr:
Hmm, ideally we would have something like a query builder for many of those things. Of course that would increase the overall code, but then things could become a lot more flexible and documented compared to free functions and raw queries.

Alexandcoats:
Yes, I agree on both counts. And I think that is too far outside the scope of the PR, which is why I did not attempt it.

Alex6323:
I lean towards @Alexandcoats here. Moving parts of the query into functions doesn't really help here. It has the disadvantages that Alex already stated. The correctness of the queries should always be guaranteed by tests. If our test suite doesn't catch that divergence you mentioned, Jochen, then that's a bug in our test suite (we missed to write a necessary test case).
Moreover, I really prefer the atomicity of our current queries meaning that all I need to read about a particular query is in one place and I don't have to jump back and forth.

PR #963, opened 2022-12-07
Proposal to simplify route management

grtlr:


src/bin/inx-chronicle/api/mod.rs
```diff
@@ -37,33 +37,59 @@ pub use self::{
 
 pub const DEFAULT_PAGE_SIZE: usize = 100;
 
+#[derive(Clone, Debug, Default)]
+pub struct RegisteredRoutes(Vec<String>);
+
+impl RegisteredRoutes {
+    pub fn register(&mut self, route: impl Into<String>) -> String {
```
Alexandcoats, line 44: Maybe just accept a `&str` and return it to avoid the allocation weirdness.

grtlr:
Woahh, where did that come from? :trollface: What do you think about adding `ApiState` and making our own `Router` non-generic?

grtlr:
So here is the thing:

* We currently don't need a more sophisticated implementation (because we don't do JWT-based returns or different nestings, and neither does Hornet).
* No one actually uses this route for more than checking which high-level APIs are available.

So I think we should follow YAGNI here, we still have the code in the repo, if we ever want to come back to it.

grtlr:
Closing for now due to inactivity.

PR #979, opened 2022-12-13
refactor(config)!: re-design and clean-up configuration

### LINKED ISSUE #788: Chronicle should not use a hard-coded default config

### LINKED ISSUE #947: `GenerateJWT` CLI command does not seem to be generating valid tokens
### Bug description

The `generate-jwt` command does not seem to generate valid tokens, even though it is identical to the login route (which does work).

### LINKED ISSUE #966: Refactor Chronicle Configuration
## Description

Refactor Chronicle's configuration. 

## Motivation

* reduce confusion for the user about config parameter precedence;
* fix default value inconsistencies between config default impls and `config.template.toml`;
* prevent any default credentials from being committed to the repo (see also: https://12factor.net/config);
* make it easier to run Chronicle in different scenarios (different networks etc);
* document default parameters in the CLI;

Yes.

### LINKED ISSUE #1000: Load `.env` file in `docker-compose.yml`
## Description

This should streamline local vs Docker-ized development. While doing this we should also ensure that `dotenvy` integration still works.

Alex6323:
## Description and Motivation

This PR changes the way how Chronicle can be configured. This was motivated by the fact that we defined defaults in multiple locations and naturally ... conflicts appeared and bit us ... some place. Although those defaults were in theory (if they had been consistent) documented in the `config.template.toml` file, we think those are better documented in the CLI itself (if you invoke `--help`). Another problem with such a file is that users were supposed to make a copy of that file and change it to their liking. This resulted in the `config.toml` being invalidated earlier than it needed to be since often times no parameter that the user actually adapted was conflicting with the newer version. This PR changes the preferred mode of configuration for Chronicle as it makes everything configurable through the CLI itself and also limits supported environment variables to credentials basically (which are more unlikely to be deprecated). Together with the focus of Chronicle as a Docker deployment, this also allows for making our `docker-compose.yml` the main location of configuration as it's no longer needed to copy over the template file. 

## Linked Issues

* Closes #966
* Closes #947
* Closes #1000 

## Notes to Reviewer

#### Config Changes
* [x] Ensure that config changes work with individual build features by running `cargo ci-check-features`.


.github/workflows/_test_int.yml
```diff
@@ -37,7 +37,9 @@ jobs:
         uses: supercharge/mongodb-github-action@1.8.0
         with:
           mongodb-version: ${{ inputs.mongodb }}
-          mongodb-replica-set: test-rs
+          #mongodb-replica-set: test-rs
```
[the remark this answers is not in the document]
Alex6323: basically because of this issue:

src/bin/inx-chronicle/cli.rs
```diff
@@ -1,195 +1,244 @@
...
 
-        if let Some(conn_str) = &self.mongodb.mongodb_conn_str {
-            config.mongodb.conn_str = conn_str.clone();
-        }
-
-        if let Some(db_name) = &self.mongodb.mongodb_database {
-            config.mongodb.database_name = db_name.clone();
-        }
+        // MongoDb
+        // Note: all unwraps are fine because we defined defaults for all, so none of them can be None ;)
```
grtlr, line 186: Which `unwraps`?

Alex6323: haha, forgot to remove this note

src/bin/inx-chronicle/cli.rs
```diff
@@ -1,195 +1,243 @@
...
-            .config
-            .as_ref()
-            .map(ChronicleConfig::from_file)
-            .transpose()?
-            .unwrap_or_default();
+    /// Get a config from a file (specified via the `--config` option) or from provided CLI args combined
+    /// with defaults for those that are not provided. Note that a config file must be fully specified
+    /// as it cannot be overwritten with the CLI defaults. If you plan on using a `config.toml` use
+    /// Chronicle's `create-config' tool to make sure of that.
+    pub fn get_config(&self) -> ChronicleConfig {
```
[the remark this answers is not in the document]
Alex6323: good catch

grtlr (changes requested):
Please make sure `cargo ci-check-features` command passes.

src/bin/inx-chronicle/cli.rs
```diff
@@ -113,12 +113,20 @@ impl From<&InfluxDbArgs> for chronicle::db::influxdb::InfluxDbConfig {
...
+            #[cfg(not(feature = "analytics"))]
+            analytics_database_name: String::new(),
             #[cfg(feature = "metrics")]
             metrics_enabled: !value.disable_metrics,
+            #[cfg(not(feature = "metrics"))]
+            metrics_enabled: false,
             #[cfg(feature = "metrics")]
             metrics_database_name: value.metrics_database_name.clone(),
+            #[cfg(not(feature = "metrics"))]
+            metrics_database_name: String::new(),
```
grtlr, line 129: Shouldn't those be feature-gated in `InfluxDbConfig` too so that this step would not be necessary?

grtlr (changes requested):

grtlr:
Could you please also do https://github.com/iotaledger/inx-chronicle/issues/1000 as part of this PR? 🙏

src/db/mongodb/collection.rs
```diff
@@ -145,12 +145,13 @@ pub trait MongoDbCollectionExt: MongoDbCollection {
 }
 impl<T: MongoDbCollection> MongoDbCollectionExt for T {}
 
-pub(crate) struct InsertResult {
-    pub(crate) _ignored: usize,
+pub struct InsertResult {
+    pub _ignored: usize,
```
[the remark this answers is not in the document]
Alex6323: Can we solve this when we deal with the visibility modifiers. This has nothing to do with this PR really. It's just to make the CI happy, and I think we should address this separatedly.

README.md
```diff
@@ -19,3 +19,16 @@ The changelog can be created using the following command (requires the [`convent
 ```sh
 conventional-changelog -p conventionalcommits -i CHANGELOG.md -s
 ```
+
+## Docker deployment configuration of credentials through environment variables
+
+Docker compose will automatically load credentials for different services from a `.env` file that must be located in the same directory as the `docker-compose.yml` file. You therefore must create such a file before you do a `docker compose up`. An example `.env` file could look like this:
```
grtlr, line 25:
```suggestion
Docker compose will automatically load credentials for different services from a `.env` file that must either be located in the same directory as the `docker-compose.yml` file, or specified using the `--env-file` flag. You therefore must create such a file before you do a `docker compose up`. An example `.env` file could look like this:
```

grtlr (approved):

PR #995, opened 2022-12-22
feat: use `Stream`s of `LedgerUpdateMessage`s as abstraction

### LINKED ISSUE #959: `fill-analytics` with `-n 10` makes per milestone time drop significantly
### Bug description

image

The sync time per milestone is higher when running with `-n 10`, also the variation in sync time is higher. One reason for this could be the access pattern for the data (interleaving).

We should investigate if we can improve this to speed up the syncing process.

### LINKED ISSUE #981: Implement accumulating metrics for both INX and regular setups
We currently have #924, which computes incremental analytics based on the querying the database. This gives us a nice speedup of 50%, but for INX-based syncs we should be able to do better by working directly on the ledger updates received via INX.

grtlr:
## Linked Issues

* Closes #981.
* Closes #959.

## Notes to Reviewer

#### Test cases

#### INX Changes

#### Database Changes


grtlr:
We continued this idea in #1049.

PR #1005, opened 2022-12-28
fix(analytics): computation of daily active addresses

grtlr:
Daily active addresses were not computed because we recreated the `Vec` of analytics for each ledger update, leading to a reset of `current_date`.

## Notes to Reviewer

#### INX Changes


Alexandcoats:
It feels like the wrong place to do the check for "is this the next day" in `get_measurement`, but maybe we can revisit the design later.

grtlr:
Yeah, I think we should overhaul our general design of analytics!

Alex6323 (approved):

PR #1025, opened 2023-01-10
chore(deps): remove `toml`

grtlr:


Alex6323 (approved):
right, we don't need that anymore!

PR #1026, opened 2023-01-10
chore(clippy): fix `Display` formatting

grtlr:
This is to satisfy future versions of `clippy`.


Alex6323 (approved):
I see what you meant about new year cleaning :trollface:

PR #1027, opened 2023-01-10
fix(ci): fix coverage workflow

grtlr:


Alex6323 (commented):
doesn't seem to work ...

PR #1028, opened 2023-01-10
fix(analytics): add `total_byte_cost` to ledger size analytics

### LINKED ISSUE #996: `total_storage_deposit_value` analytic is wrong
Our definition of `total_storage_deposit_value` is wrong:

https://github.com/iotaledger/inx-chronicle/blob/6ec481cc2200c7e7990770c2f20ace336bc0b3e7/src/db/collections/analytics/ledger_size.rs#LL78C21-L78C21

In addition to the _storage deposit return unlock condition amount_ we should also track to total required storage cost (i.e. the _total_byte_cost_).

Alexandcoats:
## Linked Issues

* Closes #996

## Notes to Reviewer

#### INX Changes

#### Database Changes


grtlr:
Can you please also update the Grafana dashboard to use the `total_byte_cost` value instead of the `total_storage_deposit_value`?

src/db/collections/analytics/ledger_size.rs
```diff
@@ -64,30 +59,62 @@ impl OutputCollection {
...
+            total_key_bytes: res.total_key_bytes,
+            total_data_bytes: res.total_data_bytes,
+            total_byte_cost: res
+                .rent_structure
+                .map(|rs| {
+                    d128::from(rs.v_byte_cost)
+                        * ((res.total_key_bytes * d128::from(rs.v_byte_factor_key as u32))
+                            + (res.total_data_bytes * d128::from(rs.v_byte_factor_data as u32)))
+                })
+                .unwrap_or_default(),
```
grtlr, line 117: I think we should rather `impl iota_types::block::output::Rent for ::Result` and then call `.rent_cost(...)`.

src/db/collections/analytics/ledger_size.rs
```diff
@@ -25,15 +28,7 @@ pub struct LedgerSizeAnalyticsResult {
...
-}
-
-impl LedgerSizeAnalyticsResult {
-    pub fn total_byte_cost(&self, protocol_params: &ProtocolParameters) -> d128 {
-        let rent_structure = protocol_params.rent_structure;
-        d128::from(rent_structure.v_byte_cost)
-            * ((self.total_key_bytes * d128::from(rent_structure.v_byte_factor_key as u32))
-                + (self.total_data_bytes * d128::from(rent_structure.v_byte_factor_data as u32)))
-    }
+    pub total_byte_cost: d128,
```
grtlr, line 31: These should just use `u64`. The total byte cost can't exceed the number of tokens in the network. Same goes for the other fields.

Alexandcoats: The problem is there is no way to convert `d128` to `u64` in the lib, and we have to use `Decimal128` in the query to ensure we have enough space. The only option I found is to convert through a string. Not sure it's worth it since we're just going to convert back to a string for influx.

grtlr: Ah yeah, makes sense.

src/db/collections/analytics/ledger_size.rs
```diff
@@ -64,30 +60,63 @@ impl OutputCollection {
...
+                    doc! { "$group" : {
+                        "_id": null,
+                        "total_key_bytes": { "$sum": { "$toDecimal": "$details.rent_structure.num_key_bytes" } },
+                        "total_data_bytes": { "$sum": { "$toDecimal": "$details.rent_structure.num_data_bytes" } },
+                        "total_storage_deposit_value": { "$sum": { "$toDecimal": "$output.storage_deposit_return_unlock_condition.amount" } },
+                    } },
+                    doc! { "$lookup": {
+                        "from": ProtocolUpdateCollection::NAME,
+                        "pipeline": [
+                            { "$match": { "_id": { "$lte": ledger_index } } },
```
Alex6323, line 87: This would return the first it finds that is `< ledger_index`, but not necessarily the closest, right? Feels like this could become a problem.

Alexandcoats: Hm. Good catch.

src/db/collections/analytics/ledger_size.rs
```diff
@@ -64,30 +60,64 @@ impl OutputCollection {
...
+                vec![
+                    doc! { "$match": {
+                        "metadata.booked.milestone_index": { "$lte": ledger_index },
+                        "metadata.spent_metadata.spent.milestone_index": { "$not": { "$lte": ledger_index } }
+                    } },
+                    doc! { "$group" : {
+                        "_id": null,
+                        "total_key_bytes": { "$sum": { "$toDecimal": "$details.rent_structure.num_key_bytes" } },
+                        "total_data_bytes": { "$sum": { "$toDecimal": "$details.rent_structure.num_data_bytes" } },
+                        "total_storage_deposit_value": { "$sum": { "$toDecimal": "$output.storage_deposit_return_unlock_condition.amount" } },
```
grtlr, line 82: This should be removed, as it's already part of the `UnlockConditionAnalytics`.

src/db/collections/analytics/influx.rs
```diff
@@ -58,18 +58,10 @@ impl From<Measurement> for influxdb::WriteQuery {
...
-                    m.inner.total_key_bytes.to_string().parse::<u64>().unwrap(),
-                )
-                .add_field(
-                    "total_data_bytes",
-                    m.inner.total_data_bytes.to_string().parse::<u64>().unwrap(),
-                ),
+                .add_field("total_storage_deposit_value", m.inner.total_storage_deposit_value)
+                .add_field("total_key_bytes", m.inner.total_key_bytes)
+                .add_field("total_data_bytes", m.inner.total_data_bytes)
+                .add_field("total_byte_cost", m.inner.total_byte_cost),
```
grtlr, line 64: Rename this to `total_storage_deposit_value` and remove the previous statistic based on the unlock condition.

Alexandcoats: Like, remove it from everywhere?

grtlr: Yes and replace it with `m.inner.total_byte_cost` please.

grtlr (approved):

src/db/collections/analytics/unlock_condition.rs
```diff
@@ -90,19 +91,53 @@ impl OutputCollection {
...
+            count: u64,
+            value: d128,
+            inner: d128,
+        }
+
+        let sdruc = 
+            self.aggregate::<ResSdruc>(
+                vec![
+                    doc! { "$match": {
+                        format!("output.storage_deposit_return_unlock_condition"): { "$exists": true },
```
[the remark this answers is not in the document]
grtlr: Accidental

PR #1030, opened 2023-01-11
chore(docker): clean up unused variables and files

grtlr:


Alexandcoats:
These are unused?

grtlr:
The InfluxDB stuff is actually harmful and comes from a time where we pushed to many series (because we had to many distinct tags). So I wanted to revert that. Since the Hornet nest does not contain value transactions it's kinda useless for us. I've personally never used it. Note that we can still treat it just like the private tangle.

Alex6323 (approved):

PR #1041, opened 2023-01-17
feat(analytics): add milestone byte size analytic

### LINKED ISSUE #1007: Add data throughput analytic at block level
We should gather analytics on the throughput of the Tangle by counting the bytes of the raw blocks that we receive per milestone.

Alex6323:
## Linked Issues

* Closes #1007 

## Notes to Reviewer

#### Database Changes


src/db/collections/analytics/influx.rs
```diff
@@ -61,6 +61,9 @@ impl From<Measurement> for influxdb::WriteQuery {
                 .add_field("total_key_bytes", m.inner.total_key_bytes)
                 .add_field("total_data_bytes", m.inner.total_data_bytes)
                 .add_field("total_storage_deposit_value", m.inner.total_storage_deposit_value),
+            Measurement::MilestoneSizeAnalytics(m) => m
+                .prepare_query("stardust_milestone_size")
+                .add_field("total_milestone_bytes", m.inner.total_milestone_bytes),
```
grtlr, line 66: Would it make sense to compute this per payload type? Both tagged data payloads and outputs allow adding data to the tangle.

grtlr (commented):
Thanks for opening this PR! 

Can you please also add a Grafana panel once you're done? 🙏

Alexandcoats:
This seems still very incomplete (no dashboard, no CLI support), so it should be a draft, no?

Alex6323:
you're right.. not ready for review yet so should have been a draft

src/db/collections/analytics/influx.rs
Alex6323: We could certainly do that, but the question is whether there's something interesting to draw conclusions from that analytic. Can you come up with something? I think I'ld not add this preemptively but rather change it when there's an actual request/demand for it.

grtlr: I think it's interesting to see. This way we could for example try to gauge if people start using the ledger to store more and more information (remember we also push number of payloads, so we can compute averages in InfluxDb).

grtlr: Also, if we change it later we have to resync.

Alex6323: right, okay I'll add it.

grtlr:
Closing in favor of #1061, which has the same functionality.

PR #1050, opened 2023-01-18
Restructuring tangle abstraction and fixing errors

grtlr:


src/tangle/cone_stream.rs
```diff
@@ -0,0 +1,60 @@
...
+use pin_project::pin_project;
+
+use super::{ledger_updates::LedgerUpdateStore, InputSource};
+use crate::types::{
+    ledger::BlockMetadata,
+    stardust::block::{payload::TransactionEssence, Block, BlockId, Input, Output, Payload},
+};
+
+/// A [`Block`] enriched with [`BlockMetadata`] and potentially [`Output`]s that correspond to inputs.
+pub struct BlockWithMetadataInputs {
```
Alex6323, line 19: not very constructive feedback I know but those more and more frankenstein'ed struct names indicate that we're not modelling this right IMO.

Alex6323: The problem is that we're mixing two layers here: the `Block` layer and the `Transaction` layer (that builds on top of the `Block` layer), and I am not sure we should do that.

Alex6323: Wouldn't it be better to have for example a `TransactionStream` type, that only streams all `Transactions` with their inputs and outputs and relevant metadata as they occured within a milestone?

PR #1051, opened 2023-01-18
Add `MilestoneRange`

Alexandcoats:


src/tangle/sources/mod.rs
```diff
@@ -28,14 +72,14 @@ pub trait InputSource {
...
         &self,
-        range: std::ops::Range<MilestoneIndex>,
+        range: MilestoneRange,
     ) -> Result<BoxStream<Result<MilestoneAndProtocolParameters, Self::Error>>, Self::Error>;
 
     /// Retrieves a stream of blocks and their metadata in white-flag order given a milestone index.
     async fn cone_stream(
         &self,
         index: MilestoneIndex,
-    ) -> /* TODO: This should not require enriching the inputs already */ Result<BoxStream<Result<BlockWithMetadataInputs, Self::Error>>, Self::Error>;
```
grtlr, line 38: That `TODO` should probably stay there.

src/tangle/sources/mod.rs
```diff
@@ -19,6 +23,46 @@ pub struct MilestoneAndProtocolParameters {
...
+        if self.start <= self.end {
+            let res = self.start;
+            self.start += 1;
+            Some(res)
+        } else {
+            None
+        }
+    }
+}
+
```
grtlr, line 65: I feel like this should go into either the `milestone_stream.rs` or a separate `milestone_range.rs` file.

src/tangle/sources/mod.rs
Alexandcoats: It got removed automagically, but i re-added it.

grtlr (approved):

PR #1054, opened 2023-01-18
Add impls for INX and in-memory data

Alexandcoats:


src/tangle/sources/inx.rs
```diff
@@ -1,2 +1,90 @@
...
+#[async_trait]
+impl InputSource for Inx {
+    type Error = InxError;
+
+    async fn milestone_stream(
+        &self,
+        range: MilestoneRange,
+    ) -> Result<BoxStream<Result<MilestoneData, Self::Error>>, Self::Error> {
+        let mut inx = self.clone();
+        Ok(Box::pin(
```
grtlr, line 28: [nitpick] It's maybe a matter of personal perference, but you can also call `.boxed()` on the stream.

src/tangle/sources/inx.rs
```diff
@@ -1,2 +1,90 @@
...
+            inx.listen_to_confirmed_milestones(range.into())
+                .await?
+                .and_then(|msg| async move {
+                    let payload = if let iota_types::block::payload::Payload::Milestone(payload) =
+                        msg.milestone.milestone.inner_unverified()?
+                    {
+                        payload.into()
+                    } else {
+                        // The raw data is guaranteed to contain a milestone payload.
+                        unreachable!();
```
grtlr, line 38: I think I prefer putting the documentation into the panic like so:
```suggestion
                        unreachable!("Raw milestone data has to contain a milestone payload");
```

src/tangle/sources/inx.rs
```diff
@@ -1,2 +1,90 @@
...
+    }
+
+    async fn ledger_updates(&self, index: MilestoneIndex) -> Result<LedgerUpdateStore, Self::Error> {
+        let mut inx = self.clone();
+        let mut stream = inx.listen_to_ledger_updates((index.0..=index.0).into()).await?;
+        let MarkerMessage {
+            consumed_count,
+            ..
+            // TODO: What do we do here?
+        } = stream.try_next().await?.unwrap().begin().unwrap();
```
grtlr, line 78: I think we should rip everything out of the `inx` module and put it into this module / impl here. We should then extend the `InxError` type with the variants that are required by our new implementation.

Alexandcoats: Alright, maybe we should create an issue then to track that progress.

src/tangle/sources/inx.rs
Alexandcoats: `Box::pin` does not require `futures::StreamExt`

src/tangle/sources/inx.rs
grtlr: Sounds good, can you open the issue please?

grtlr (approved):

PR #1056, opened 2023-01-19
Impl ledger queries for MongoDb input source

Alex6323:


Alexandcoats (approved):

src/db/collections/outputs/mod.rs
```diff
@@ -277,6 +277,55 @@ impl OutputCollection {
...
+                    "rent_structure": "$details.rent_structure",
+                } },
+            ],
+            None,
+        )
+        .await
+    }
+
+    /// Get all ledger updates (i.e. consumed [`Output`]s) for the given milestone.
+    pub async fn get_ledger_updates(&self, ledger_index: MilestoneIndex) -> Result<Vec<(OutputId, Output)>, Error> {
```
grtlr, line 304: Is there a reason why we collect into a `Vec`? If we return an iterator, we could save this allocation. We already do that for `get_unspent_output_stream`.

Alex6323: I think for this particular data `(OutputId, Output)` it's save to directly collect in to a `HashMap`, which then can be moved.

grtlr (approved):

Issue #1048, opened 2023-01-17
Implement analytics on top of `InputSource` abstraction


grtlr:
Should be synchronous in the beginning, we can split them into tasks later.

We should start with the prototypical analytics first (for the different use cases first):

* [x] Addresses with balance (ledger state)
* [x] Payload types (tangle level)
* [x] Ledger size (#1060)

Keep _daily active addresses_ in mind.

Alexandcoats:
Closed by #1061

PR #1060, opened 2023-01-23
feat(analytics): impl ledger size analytics with `InputSource` abstraction

Alex6323:
See #1048

### DIFF
--- a/src/db/collections/analytics/ledger_size.rs
+++ b/src/db/collections/analytics/ledger_size.rs
@@ -2,116 +2,181 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use async_trait::async_trait;
-use futures::TryStreamExt;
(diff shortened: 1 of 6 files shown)


src/db/collections/analytics/ledger_size.rs
```diff
@@ -2,116 +2,181 @@
...
+                    }
+                }
+                _ => (),
+            }
+        }
+        if let Some(inputs) = inputs {
+            for input in inputs {
+                self.stats -= input.ledger_stat(protocol_params);
+            }
+        }
```
grtlr, line 74: Reminder: Check if a TX was actually applied

Alex6323: thanks, you're absolutely right of course :+1:

grtlr:
This is superseded by #1070.

PR #1061, opened 2023-01-23
Wip

grtlr:


src/analytics/address_balance.rs
```diff
@@ -0,0 +1,38 @@
...
+        }
+
+        for output in outputs {
+            if let Some(a) =  output.owning_address() {
+                self.addresses.insert(a.clone());
+            }
+        }
+    }
+
+    fn end_milestone(&mut self, index: MilestoneIndex) -> Option<Self::Measurement> {
```
Alex6323, line 35: Why the `Option`? Isn't it safe to assume that whenever `end_milestone` is called the measurement has been created?

src/analytics/block_activity.rs
```diff
@@ -0,0 +1,44 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+use super::BlockAnalytics;
+use crate::types::{
+    ledger::BlockMetadata,
+    stardust::block::{Block, Output, Payload},
+    tangle::MilestoneIndex,
+};
+
+#[derive(Clone, Debug, Default)]
```
Alex6323, line 11: should be fine to make this type also `Copy`

src/analytics/mod.rs
```diff
@@ -0,0 +1,27 @@
...
+};
+
+pub mod address_balance;
+pub mod block_activity;
+
+#[allow(missing_docs)]
+pub trait BlockAnalytics {
+    type Measurement;
+    fn begin_milestone(&mut self, index: MilestoneIndex);
+    fn handle_block(&mut self, block: &Block, block_metadata: &BlockMetadata, inputs: &Option<Vec<Output>>);
```
Alex6323, line 17: shouldn't `inputs` rather be `Option<&[Output]>`?

src/analytics/address_balance.rs
Alex6323: Oh, I think I got it ... the idea is to skip intermediate incremental states in which we're not interested in.

src/analytics/mod.rs
grtlr: We might not even need the `inputs` here at all. Everything related to the ledger should probably be handled by `TransactionAnalytics`.

src/analytics/base_token.rs
```diff
@@ -0,0 +1,54 @@
...
+    fn begin_milestone(&mut self, _: MilestoneIndex) {
+        self.measurement = BaseTokenActivity::default();
+    }
+
+    fn handle_transaction(&mut self, inputs: &[LedgerSpent], outputs: &[LedgerOutput]) {
+        let mut outflows: HashMap<&Address, usize> = HashMap::new();
+        for input in inputs {
+            if let Some(address) = input.output.output.owning_address() {
+                self.measurement.transferred_value += input.output.output.amount().0 as usize;
+                *outflows.entry(address).or_default() += input.output.output.amount().0 as usize;
```
grtlr, line 38: We should probably define helper functions to avoid `output.output.amount` weirdness.

src/analytics/mod.rs
Alex6323: Makes sense.

src/analytics/base_token.rs
Alex6323: I agree. That or maybe we consider refactoring to `inner`. I think reading `input.output.inner.amount` would be fine. (reading `input.output.....` is also a bit weird I guess :smile:

Alexandcoats: Obviously we should change `input` to `output` so that it's `output.output.output.amount`.

src/analytics/ledger/active_addresses.rs
```diff
@@ -0,0 +1,79 @@
...
+
+use time::{Duration, OffsetDateTime};
+
+use super::{AddressCount, TransactionAnalytics};
+use crate::types::{
+    ledger::{LedgerOutput, LedgerSpent, MilestoneIndexTimestamp},
+    stardust::block::Address,
+};
+
+/// Computes the number of addresses the currently hold a balance.
```
Alexandcoats, line 14: Inaccurate comment, no?

src/analytics/ledger/address_balance.rs
```diff
@@ -3,28 +3,24 @@
...
     stardust::block::Address,
-    tangle::MilestoneIndex,
 };
 
-/// The number of addresses.
-pub struct AddressCount(usize);
-
 /// Computes the number of addresses the currently hold a balance.
 pub struct AddressBalanceAnalytics {
     addresses: HashSet<Address>,
```
Alexandcoats, line 14: This is definitely not right. There's no way to know if they hold a balance without holding said balance in the map and subtracting out transaction inputs.

src/analytics/ledger/base_token.rs
```diff
@@ -27,7 +26,7 @@ pub struct BaseTokenActivityAnalytics {
 impl TransactionAnalytics for BaseTokenActivityAnalytics {
     type Measurement = BaseTokenActivity;
 
-    fn begin_milestone(&mut self, _: MilestoneIndex) {
+    fn begin_milestone(&mut self, _: MilestoneIndexTimestamp) {
         self.measurement = BaseTokenActivity::default();
     }
```
Alexandcoats, line 32: Shouldn't we only count transferred value if the input doesn't have the same address as the outputs?

src/analytics/ledger/mod.rs
```diff
@@ -4,24 +4,30 @@
...
+    outputs::{CountValue, OutputState, OutputStatistics},
     unclaimed_tokens::{UnclaimedTokens, UnclaimedTokensAnalytics},
 };
 
 #[allow(missing_docs)]
 pub trait TransactionAnalytics {
     type Measurement;
-    fn begin_milestone(&mut self, index: MilestoneIndex);
+    fn begin_milestone(&mut self, at: MilestoneIndexTimestamp);
     fn handle_transaction(&mut self, inputs: &[LedgerSpent], outputs: &[LedgerOutput]);
```
Alexandcoats, line 28: Maybe we should rename these to created and consumed.

src/analytics/ledger/outputs.rs
```diff
@@ -0,0 +1,96 @@
...
+            match input.output.output {
+                Output::Alias(_) => self.measurement.alias -= input,
+                Output::Basic(_) => self.measurement.basic -= input,
+                Output::Nft(_) => self.measurement.nft -= input,
+                Output::Foundry(_) => self.measurement.foundry -= input,
+                Output::Treasury(_) => self.measurement.treasury -= input,
+            }
+        }
+
+        for output in outputs {
```
Alexandcoats, line 82: Add first to avoid underflow

src/analytics/ledger/address_balance.rs
```diff
@@ -0,0 +1,52 @@
...
+
+impl TransactionAnalytics for AddressBalanceAnalytics {
+    type Measurement = AddressCount;
+
+    fn begin_milestone(&mut self, _: MilestoneIndexTimestamp) {}
+
+    fn handle_transaction(&mut self, inputs: &[LedgerSpent], outputs: &[LedgerOutput]) {
+        for input in inputs {
+            if let Some(a) = input.output.output.owning_address() {
+                self.addresses.remove(a);
```
shufps, line 38: not sure if this is right because addresses are not consumed 100% like UTXOs. You can have an address and send coins away and still have balance on the address.

src/analytics/ledger/outputs.rs
grtlr: I thought about this some more: The underflow should never happen because of the `init` call as we can only spent an output that was actually present in `read_unspent_outputs`. So if this underflows, we have a problem in our underlying logic, which is why I'm leaving this in for now.

src/analytics/block_activity.rs
grtlr: I think for now I like the explicitness of the copy until we have settled on a good API for returning the measurements. Performance-wise it shouldn't make a difference.

src/analytics/ledger/base_token.rs
shufps: you are right, this implementation looks wrong :thinking: 

for comparison: https://github.com/iotaledger/inx-analytics/blob/develop/pkg/analytics/ledger_output_analytics.go#L250-L264

src/analytics/ledger/address_balance.rs
shufps: for comparison: https://github.com/iotaledger/inx-analytics/blob/develop/pkg/analytics/ledger_output_analytics.go#L358-L381

shufps: for comparison: https://github.com/iotaledger/inx-analytics/blob/develop/pkg/analytics/ledger_output_analytics.go#L358-L381

src/analytics/ledger/base_token.rs
grtlr: Thanks for pointing this out! Should be fixed now.

src/analytics/ledger/address_balance.rs
```diff
@@ -0,0 +1,59 @@
...
+
+impl TransactionAnalytics for AddressBalanceAnalytics {
+    type Measurement = AddressCount;
+
+    fn begin_milestone(&mut self, _: MilestoneIndexTimestamp) {}
+
+    fn handle_transaction(&mut self, inputs: &[LedgerSpent], outputs: &[LedgerOutput]) {
+        for input in inputs {
+            if let Some(a) = input.owning_address() {
+                // All inputs should be present in `addresses`. If not, we skip it's value.
```
Alexandcoats, line 38: This seems...strange. We should never skip a value, right? I mean, perhaps we should just return an error in that case, even if it should never happen.

src/analytics/ledger/address_balance.rs
```diff
@@ -0,0 +1,59 @@
...
+                if let Some(amount) = self.addresses.get_mut(a) {
+                    *amount -= input.amount();
+                    if *amount == OutputAmount(0) {
```
Alexandcoats, line 41: Why construct a type just to compare?

src/analytics/ledger/unclaimed_tokens.rs
```diff
@@ -0,0 +1,55 @@
...
+        Self { measurement }
+    }
+}
+
+impl TransactionAnalytics for UnclaimedTokensAnalytics {
+    type Measurement = UnclaimedTokens;
+
+    fn begin_milestone(&mut self, _: MilestoneIndexTimestamp) {}
+
+    fn handle_transaction(&mut self, inputs: &[LedgerSpent], _: &[LedgerOutput]) {
```
Alexandcoats, line 43: Didn't rename all of these to created/consumed.

src/analytics/ledger/base_token.rs
grtlr: @grtlr Please add comments. :trollface:

src/analytics/ledger/ledger_size.rs
```diff
@@ -0,0 +1,76 @@
...
+    stardust::block::Output,
+    tangle::ProtocolParameters,
+};
+trait LedgerSize {
+    fn ledger_size(&self, protocol_params: &ProtocolParameters) -> LedgerSizeMeasurement;
+}
+
+impl LedgerSize for Output {
+    fn ledger_size(&self, protocol_params: &ProtocolParameters) -> LedgerSizeMeasurement {
+        // Unwrap: acceptable risk
```
Alexandcoats, line 19: https://www.youtube.com/watch?v=TPpUafsSSaY

Alexandcoats (approved):

src/analytics/ledger/base_token.rs
```diff
@@ -0,0 +1,60 @@
...
+        self.booked_amount = TokenAmount(balance_deltas.values().copied().sum::<i128>() as u64);
+
+        // Afterwards, we subtract the tokens from that address to get the actual deltas of each account.
+        for output in consumed {
+            if let Some(address) = output.owning_address() {
+                *balance_deltas.entry(address).or_default() -= output.amount().0 as i128;
+            }
+        }
+
+        // The number of transferred tokens is then the sum of all deltas.
```
Alex6323, line 46:
```suggestion
        // The number of transferred tokens is then the sum of all positive deltas.
```

grtlr: Nah, it's important for correct analytics. Since transferring tokens is essentially a zero-sum game, the positive and negative deltas would cancel out (unless tokens are burned of course).

PR #1065, opened 2023-01-23
fix(analytics): only calculate analytics after chronicle is synced

Alexandcoats:
## Notes to Reviewer

#### INX Changes
* [ ] Ensure that restarting does not change the start milestone.

#### Database Changes

### DIFF
--- a/src/bin/inx-chronicle/stardust_inx/mod.rs
+++ b/src/bin/inx-chronicle/stardust_inx/mod.rs
@@ -314,14 +314,22 @@ impl InxWorker {
 
             let starting_index = starting_index.unwrap_or_default();
 
-            info!("Setting starting index to {}", starting_index);
-
             // Get the timestamp for the starting index
             let milestone_timestamp = inx
                 .read_milestone(starting_index.into())
                 .await?
                 .milestone_info
                 .milestone_timestamp;
+
+            info!(
+                "Setting starting index to {} with timestamp {}",
+                starting_index,
+                time::OffsetDateTime::from_unix_timestamp(milestone_timestamp as _)
+                    .unwrap()
+                    .format(&time::format_description::well_known::Rfc3339)
+                    .unwrap()
+            );
+
             let starting_index = starting_index.with_timestamp(milestone_timestamp.into());
 
             self.db
--- a/src/bin/inx-chronicle/stardust_inx/mod.rs
+++ b/src/bin/inx-chronicle/stardust_inx/mod.rs
@@ -273,9 +273,26 @@ impl InxWorker {
                 .instrument(trace_span!("inx_read_unspent_outputs"))
                 .await?;
 
+            let mut starting_index = None;
+
             let mut count = 0;
             let mut tasks = unspent_output_stream
-                .inspect(|_| count += 1)
+                .inspect_ok(|_| count += 1)
+                .map(|msg| {
+                    let msg = msg?;
+                    let ledger_index = &msg.ledger_index;
+                    if let Some(index) = starting_index.as_ref() {
+                        if index != ledger_index {
+                            bail!(InxWorkerError::InvalidUnspentOutputIndex {
+                                found: *ledger_index,
+                                expected: *index,
+                            })
+                        }
+                    } else {
+                        starting_index = Some(*ledger_index);
+                    }
+                    Ok(msg)
+                })
                 .map(|res| Ok(res?.output))
                 .try_chunks(INSERT_BATCH_SIZE)
                 // We only care if we had an error, so discard the other data
@@ -295,6 +312,23 @@ impl InxWorker {
 
             info!("Inserted {} unspent outputs.", count);
 
+            let starting_index = starting_index.unwrap_or_default();
+
+            info!("Setting starting index to {}", starting_index);
+
+            // Get the timestamp for the starting index
+            let milestone_timestamp = inx
+                .read_milestone(starting_index.into())
+                .await?
+                .milestone_info
+                .milestone_timestamp;
+            let starting_index = starting_index.with_timestamp(milestone_timestamp.into());
+
+            self.db
+                .collection::<ApplicationStateCollection>()
+                .set_starting_index(starting_index)
+                .await?;
+
             info!(
                 "Linking database `{}` to network `{}`.",
                 self.db.name(),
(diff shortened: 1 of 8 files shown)


src/bin/inx-chronicle/stardust_inx/mod.rs
```diff
@@ -275,6 +298,19 @@ impl InxWorker {
                 .collection::<ProtocolUpdateCollection>()
                 .insert_protocol_parameters(start_index, protocol_parameters.into())
                 .await?;
+
+            // Get the latest milestone again after reading unspent outputs
+            let starting_index = inx
```
Alex6323, line 303: Is this index required to be the index for which we requested the unspent outputs? Or asked in a different way: if we got the unspent outputs at index A ... does it matter that this call could return some index A+x (x!=0)?

Alexandcoats: It's supposed to be, but since we have to make two calls, it's possible for it to be the next milestone, which is better than it being the previous one.

src/bin/inx-chronicle/stardust_inx/mod.rs
```diff
@@ -108,6 +108,14 @@ impl InxWorker {
     pub async fn run(&mut self) -> Result<()> {
         let (start_index, mut inx) = self.init().await?;
 
+        #[cfg(feature = "analytics")]
```
grtlr, line 111: I don't think we should feature gate this.

src/db/collections/application_state.rs
```diff
@@ -0,0 +1,57 @@
...
+        mongodb::{MongoDbCollection, MongoDbCollectionExt},
+        MongoDb,
+    },
+    types::tangle::MilestoneIndex,
+};
+
+/// The MongoDb document representation of singleton Application State.
+#[derive(Clone, Debug, Eq, PartialEq, Serialize, Deserialize)]
+pub struct ApplicationStateDocument {
+    pub starting_index: MilestoneIndex,
```
grtlr, line 18: Maybe we should at the `MilestoneTimestamp` as well if we want to use the state in other areas as well?

src/bin/inx-chronicle/stardust_inx/mod.rs
Alexandcoats: The alternative is using ignore tags. 🤷

src/db/collections/application_state.rs
Alexandcoats: So far we have no use case for that, do we?

grtlr: It would still be nice for debugging purposes even if we don't expose it yet.

src/bin/inx-chronicle/stardust_inx/mod.rs
grtlr: What I meant is, we should just always write this to the DB.

src/db/collections/application_state.rs
Alexandcoats: Alright, I'll add it.

src/bin/inx-chronicle/stardust_inx/mod.rs
Alexandcoats: That's not the issue. The problem is we only *read* the state if `analytics` is enabled.

src/bin/inx-chronicle/stardust_inx/mod.rs
```diff
@@ -235,6 +245,26 @@ impl InxWorker {
                     .insert_protocol_parameters(start_index, protocol_parameters)
                     .await?;
             }
+
+            // TODO: This is for migration purposes only. Remove it in the next version release.
```
grtlr, line 249: I think we should be able to use `#[deprecated(since = "1.0.0-beta.32")` here to enforce this.

grtlr (dismissed):

src/bin/inx-chronicle/stardust_inx/mod.rs
grtlr: Why don't we use the index returned by `read_unspent_outputs`? Isn't that guaranteed to be the correct one?

src/bin/inx-chronicle/stardust_inx/mod.rs
Alexandcoats: This would cause a warning we would also need to ignore. Also, it's the wrong usage.

Alexandcoats: I created an issue instead.

src/bin/inx-chronicle/stardust_inx/mod.rs
Alexandcoats: Depends. Is the `ledger_index` on `UnspentOutputMessage` the latest index, or is it the index of the output?

grtlr: It's the latest index, it tells us when the output was still unspent. The only thing we might need to check if all returned ledger indexes are the same (which they should be, because Hornet locks the ledger afaik).

Alexandcoats: It won't be very clean to extract that value from the stream but it's doable. I guess the question is: should we?

grtlr: I think so, and we should only set the starting index if `read_unspent_output` was sucessful and all returned outputs have the same ledger index.

src/bin/inx-chronicle/stardust_inx/mod.rs
```diff
@@ -314,14 +314,22 @@ impl InxWorker {
...
             let milestone_timestamp = inx
                 .read_milestone(starting_index.into())
                 .await?
                 .milestone_info
                 .milestone_timestamp;
+
+            info!(
+                "Setting starting index to {} with timestamp {}",
+                starting_index,
+                time::OffsetDateTime::from_unix_timestamp(milestone_timestamp as _)
```
grtlr, line 327: https://github.com/iotaledger/inx-chronicle/blob/1ddc9297fabd668696013314e59cb3cad7a1bfe9/src/types/stardust/milestone.rs#L28

Alexandcoats: Thanks

PR #1066, opened 2023-01-24
Add moar analytics

Alex6323:


src/lib.rs
```diff
@@ -2,8 +2,9 @@
 // SPDX-License-Identifier: Apache-2.0
 
 // Ideally, this would be handled completely by CI, but there is a bug in `petgraph` that prevents us from doing that.
-#![warn(missing_docs)]
-#![deny(unreachable_pub, private_in_public)]
+// FIXME: temporarily disabled
+// #![warn(missing_docs)]
+// #![deny(unreachable_pub, private_in_public)]
```
grtlr, line 7: Please re-enable

src/analytics/ledger/output_activity.rs
```diff
@@ -0,0 +1,204 @@
...
+            .filter_map(|ledger_spent| {
+                if let Output::Nft(nft_output) = &ledger_spent.output.output {
+                    if nft_output.nft_id == NftId::implicit() {
+                        // Convert implicit ids to explicit ids to make all nfts comparable
+                        // Unwrap: acceptable risk
+                        let output_id: iota_types::block::output::OutputId =
+                            ledger_spent.output.output_id.try_into().unwrap();
+                        let nft_id: NftId = iota_types::block::output::NftId::null()
+                            .or_from_output_id(&output_id)
+                            .into();
```
grtlr, line 48: This is some functionality that we should provide on our own types to avoid the conversion to `iota_types`.

src/analytics/ledger/output_activity.rs
```diff
@@ -0,0 +1,204 @@
...
+            .filter_map(|ledger_output| {
+                if let Output::Nft(nft_output) = &ledger_output.output {
+                    if nft_output.nft_id == NftId::implicit() {
+                        // Convert implicit ids to explicit ids to make all nfts comparable
+                        // Unwrap: acceptable risk
+                        let output_id: iota_types::block::output::OutputId =
+                            ledger_output.output_id.try_into().unwrap();
+                        let nft_id: NftId = iota_types::block::output::NftId::null()
+                            .or_from_output_id(&output_id)
+                            .into();
```
grtlr, line 70: Same as above.

src/analytics/ledger/output_activity.rs
```diff
@@ -0,0 +1,204 @@
...
+            .filter_map(|ledger_spent| {
+                if let Output::Alias(alias_output) = &ledger_spent.output.output {
+                    let alias_id = if alias_output.alias_id == AliasId::implicit() {
+                        // Convert implicit ids to explicit ids to make all aliases comparable
+                        // Unwrap: acceptable risk
+                        let output_id: iota_types::block::output::OutputId =
+                            ledger_spent.output.output_id.try_into().unwrap();
+                        iota_types::block::output::AliasId::null()
+                            .or_from_output_id(&output_id)
+                            .into()
```
grtlr, line 140: Same as above.

src/analytics/ledger/output_activity.rs
```diff
@@ -0,0 +1,204 @@
...
+            .filter_map(|ledger_output| {
+                if let Output::Alias(alias_output) = &ledger_output.output {
+                    let alias_id = if alias_output.alias_id == AliasId::implicit() {
+                        // Convert implicit ids to explicit ids to make all aliases comparable
+                        // Unwrap: acceptable risk
+                        let output_id: iota_types::block::output::OutputId =
+                            ledger_output.output_id.try_into().unwrap();
+                        iota_types::block::output::AliasId::null()
+                            .or_from_output_id(&output_id)
+                            .into()
```
grtlr, line 166: Same as above.

src/analytics/ledger/ledger_size.rs
```diff
@@ -0,0 +1,75 @@
...
+}
+
+pub struct LedgerSizeAnalytics {
+    protocol_params: ProtocolParameters,
+    measurement: LedgerSizeMeasurement,
+}
+
+impl LedgerSizeAnalytics {
+    // FIXME: temporarily allowed
+    #[allow(dead_code)]
```
grtlr, line 49: What's up with this?

src/analytics/ledger/ledger_size.rs
```diff
@@ -0,0 +1,75 @@
...
+    stardust::block::Output,
+    tangle::ProtocolParameters,
+};
+trait LedgerSize {
+    fn ledger_size(&self, protocol_params: &ProtocolParameters) -> LedgerSizeMeasurement;
+}
+
+impl LedgerSize for Output {
+    fn ledger_size(&self, protocol_params: &ProtocolParameters) -> LedgerSizeMeasurement {
+        // What do we do if this fails?
```
grtlr, line 19: Probably should `panic`, but maybe throw an `expect(...)` message.

src/analytics/ledger/output_activity.rs
```diff
@@ -34,8 +38,18 @@ impl TransactionAnalytics for NftActivityAnalytics {
         let inputs = inputs
             .iter()
             .filter_map(|ledger_spent| {
-                if matches!(ledger_spent.output.output, Output::Nft(_)) {
-                    Some(ledger_spent.output.output_id)
+                if let Output::Nft(nft_output) = &ledger_spent.output.output {
+                    if nft_output.nft_id == NftId::implicit() {
+                        // TODO: handle unwrap
+                        let output_id: iota_types::block::output::OutputId =
```
Alexandcoats, line 44: How sure are we about this logic with the implicit IDs?

src/analytics/tangle/mod.rs
```diff
@@ -6,13 +6,17 @@
...
+    block_activity::{BlockActivity, BlockActivityAnalytics},
+    milestone_size::{MilestoneSizeAnalytics, MilestoneSizeMeasurement},
+};
 
 #[allow(missing_docs)]
 pub trait BlockAnalytics {
     type Measurement;
     fn begin_milestone(&mut self, index: MilestoneIndex);
-    fn handle_block(&mut self, block: &Block, block_metadata: &BlockMetadata);
+    fn handle_block(&mut self, block: &Block, raw_block: &[u8], block_metadata: &BlockMetadata);
```
grtlr, line 20: Use `Raw<Block>` here.

src/analytics/ledger/output_activity.rs
```diff
@@ -0,0 +1,185 @@
...
+                    Some(AliasData {
+                        alias_id,
+                        governor_address: alias_output.governor_address_unlock_condition.address,
+                        state_index: alias_output.state_index,
+                    })
+                } else {
+                    None
+                }
+            })
+            .collect::<HashSet<_>>();
```
grtlr, line 140: We should use `HashMap<*Id,(Address,StateIndex)>` for `alias_inputs` and `alias_outputs`. To get the intersection we just iterate over either of them and index into the second. This would allow us to get rid of the ad-hoc implementations of `hash`, as well as the `AliasData` struct.

src/analytics/tangle/milestone_size.rs
```diff
@@ -0,0 +1,49 @@
...
+};
+
+/// Milestone size statistics.
+#[derive(Clone, Debug, Default)]
+pub struct MilestoneSizeMeasurement {
+    total_milestone_payload_bytes: u64,
+    total_tagged_data_payload_bytes: u64,
+    total_transaction_payload_bytes: u64,
+    total_treasury_transaction_payload_bytes: u64,
+    total_milestone_bytes: u64,
```
grtlr, line 18: Maybe we can use `usize`s here? If InfluxDB can't handle `usize` our writer should take care of the conversion.

Alex6323: fair enough ... I'll change it.

src/analytics/ledger/output_activity.rs
Alex6323: Do we want to ship or do we want to make things perfect out-of-the-gate? This is an implementation detail that can be changed later without breaking anything. Are you sure you want this changed?

grtlr: That's the thing, it's not an implementation detail, because the trait implementations are in the public interface.

Alex6323: I don't understand. The way the computation itself is done is not public. `AliasData` is private.

src/analytics/tangle/mod.rs
Alexandcoats: We can't. The methods take ownership so we can't use it as a reference.

src/analytics/ledger/output_activity.rs
Alex6323: @Alexandcoats any thoughts?

src/analytics/tangle/mod.rs
Alex6323: Indeed we'ld have to move the raw bytes and I kinda doubt that's what we would want to impose for that interface.

src/analytics/ledger/output_activity.rs
Alexandcoats: Am I wrong or did Jochen cede the point here?

Alex6323: yup, I didn't see the edit in time.

grtlr (approved):

PR #1070, opened 2023-01-26
Initial implementation of fill-analytics (and inx too :shrug:)

Alexandcoats:
This still has a lot of `todos` and weirdness but it's close-ish


src/analytics/ledger/address_balance.rs
```diff
@@ -16,19 +19,22 @@ pub struct AddressBalanceAnalytics {
...
+    pub fn init<'a>(unspent_outputs: impl IntoIterator<Item = &'a LedgerOutput>) -> Self {
+        Self {
+            addresses: unspent_outputs
+                .into_iter()
+                .fold(Default::default(), |mut addresses, output| {
+                    if let Some(&a) = output.owning_address() {
+                        *addresses.entry(a).or_default() += output.amount();
+                    }
+                    addresses
+                }),
```
grtlr, line 31: [nitpick] I find that much harder to read.

Alexandcoats: @Alex6323 If you agree I'll change it back.

src/analytics/mod.rs
```diff
@@ -3,5 +3,181 @@
...
+                Analytic::BlockActivity(stat) => stat.begin_milestone(self.at.milestone_index),
+                Analytic::DailyActiveAddresses(stat) => stat.begin_milestone(self.at),
+                Analytic::LedgerOutputs(stat) => stat.begin_milestone(self.at),
+                Analytic::LedgerSize => todo!(),
+                Analytic::OutputActivity => todo!(),
+                Analytic::ProtocolParameters(params) => {
+                    if params != &self.protocol_params {
+                        *params = self.protocol_params.clone();
+                        // TODO: either signal that we should write this at the end or do it now
+                        // TODO: re-init all of the analytics, what is this going to take? Can we even do it here?
```
grtlr, line 76:
```suggestion
                        // TODO: re-init all of the analytics, what is this going to take? Can we even do it here?
                        unimplemented!()
```

Alex6323:
That's all nice and such, but can't you do it in 100 LOC ... pls? We've all got only so much time y'know?

src/analytics/mod.rs
Alexandcoats: Oh, right, I forgot about this comment. It's already implemented, just not here.

src/analytics/ledger/output_activity.rs
```diff
@@ -6,33 +6,30 @@ use std::collections::HashSet;
...
-    pub created_count: u64,
-    pub transferred_count: u64,
-    pub destroyed_count: u64,
-}
-
-/// Measures the ledger Nft activity.
-#[derive(Debug)]
 pub struct NftActivityAnalytics {
-    measurement: NftActivityMeasurement,
+    created_count: u64,
```
grtlr, line 23: Similarly, we should be using `usize` here and in all other newly defined `*Analytics` structs.

src/analytics/mod.rs
```diff
@@ -3,5 +3,247 @@
...
+    AddressBalance(AddressBalanceAnalytics),
+    BaseTokenActivity(BaseTokenActivityAnalytics),
+    BlockActivity(BlockActivityAnalytics),
+    DailyActiveAddresses(AddressActivity),
+    LedgerOutputs(LedgerOutputAnalytics),
+    LedgerSize(LedgerSizeAnalytics),
+    MilestoneSize(MilestoneSizeAnalytics),
+    OutputActivity {
+        nft: NftActivityAnalytics,
+        alias: AliasActivityAnalytics,
```
grtlr, line 43: Now that we have in-memory analytics, we should be also able to compute `FoundryActivityAnalytics` (which we could not do before). Feel free to open an issue if you don't want to do it in this PR though.

src/bin/inx-chronicle/cli.rs
```diff
@@ -281,58 +292,94 @@ impl ClArgs {
...
-                                {
-                                    #[cfg(feature = "metrics")]
-                                    let start_time = std::time::Instant::now();
+                            let mut state: Option<crate::stardust_inx::AnalyticsState> = None;
+
+                            let tangle = Tangle::from_mongodb(&db);
+                            let mut milestone_stream = tangle
+                                .milestone_stream(start_milestone..start_milestone + chunk_size)
+                                .await?;
+                            use futures::TryStreamExt;
```
grtlr, line 317: Let's move the `fill-analytics` logic to a separate file then we also don't have to do these `use` shenanigans.

src/db/collections/analytics/address_balance.rs
```diff
@@ -1,47 +1,22 @@
...
-                Some(Measurement::AddressAnalytics(PerMilestone {
-                    milestone_index,
-                    milestone_timestamp,
-                    inner: measurement,
-                }))
-            })
-    }
-}
-
 impl OutputCollection {
```
grtlr, line 20: We should think about removing the old analytics queries/impl because they won't be used anymore. I'm afraid that both implementations will drift apart over time and introduce weird bugs. @Alex6323 wdyt?

src/analytics/ledger/active_addresses.rs
```diff
@@ -74,6 +79,8 @@ impl TransactionAnalytics for ActiveAddresses {
     }
 
     fn end_milestone(&mut self, _: MilestoneIndexTimestamp) -> Option<Self::Measurement> {
-        self.flush.take().map(AddressCount)
+        self.flush
+            .take()
+            .map(|count| DailyActiveAddressAnalyticsResult { count: count as _ })
```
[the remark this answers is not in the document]
Alexandcoats: That's fine, I just didn't want to change all of the `Measurement` impl if I didn't have to. Should I do that in this PR or wait for the next one?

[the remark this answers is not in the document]
Alexandcoats: #1080

src/db/collections/analytics/address_balance.rs
Alexandcoats: #1080

src/analytics/mod.rs
Alexandcoats: #1081

src/db/collections/analytics/base_token.rs
```diff
@@ -1,47 +1,22 @@
...
-            .map(|measurement| {
-                Some(Measurement::BaseTokenActivityAnalytics(PerMilestone {
-                    milestone_index,
-                    milestone_timestamp,
-                    inner: measurement,
-                }))
-            })
-    }
+    pub booked_value: usize,
+    pub transferred_value: usize,
```
Alex6323, line 19: Are we sure there're no issues with `usize`<->`d128` conversions?

src/db/collections/analytics/base_token.rs
```diff
@@ -51,8 +26,22 @@ impl OutputCollection {
...
+        struct Res {
+            booked_value: d128,
+            transferred_value: d128,
+        }
+
+        impl From<Res> for BaseTokenActivityAnalyticsResult {
+            fn from(value: Res) -> Self {
+                Self {
+                    booked_value: value.booked_value.to_string().parse().unwrap(),
+                    transferred_value: value.transferred_value.to_string().parse().unwrap(),
```
Alex6323, line 39: reminder to document the unwraps

src/analytics/ledger/ledger_outputs.rs
```diff
@@ -0,0 +1,106 @@
...
+    pub alias: CountValue,
+    pub basic: CountValue,
+    pub nft: CountValue,
+    pub foundry: CountValue,
+    pub treasury: CountValue,
+}
+
+impl LedgerOutputAnalytics {
+    /// Initialize the analytics by reading the current ledger state.
+    pub fn init<'a>(unspent_outputs: impl IntoIterator<Item = &'a LedgerOutput>) -> Self {
```
Alex6323, line 50: I am taking this as the example ... I noticed that many analytics need an `init`ialization which begs the question whether we should make it part of the contract itself. I know it would require another associated type to account for the possibly different initialization data that is required, but I feel like that would be cleaner and force the implementor to think about whether he has prepared the analytics object correctly for the analytic to be correct. I am fine with the impl as it is right now, but just maybe something we want to revisit later.

src/db/collections/analytics/output_activity.rs
```diff
@@ -41,39 +34,21 @@ pub struct NftActivityAnalyticsResult {
 }
 
 #[derive(Clone, Debug, Default, PartialEq, Eq, Serialize, Deserialize)]
+#[allow(missing_docs)]
+#[serde(default)]
+pub struct FoundryActivityAnalyticsResult {
+    pub created_count: u64,
+    pub transferred_count: u64,
+    pub destroyed_count: u64,
```
Alex6323, line 42: I am confused. Didn't we want to represent all counts as `usize` and only values as `u64`?

src/db/collections/analytics/milestone_size.rs
```diff
@@ -0,0 +1,12 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+#[derive(Clone, Debug, Default)]
+#[allow(missing_docs)]
+pub struct MilestoneSizeAnalyticsResult {
+    pub total_milestone_payload_bytes: u64,
+    pub total_tagged_data_payload_bytes: u64,
+    pub total_transaction_payload_bytes: u64,
+    pub total_treasury_transaction_payload_bytes: u64,
+    pub total_milestone_bytes: u64,
```
Alex6323, line 11: why `u64`? I thought we'ld use `u64` only for values and `usize` for counts?

src/db/collections/analytics/base_token.rs
grtlr: Good question. Where do we convert between these two?

src/db/collections/analytics/unlock_condition.rs
```diff
@@ -135,12 +109,12 @@ impl OutputCollection {
...
-            timelock_value: timelock.value,
+            timelock_value: timelock.value.to_string().parse().unwrap(),
             expiration_count: expiration.count,
-            expiration_value: expiration.value,
+            expiration_value: expiration.value.to_string().parse().unwrap(),
             storage_deposit_return_count: sdruc.count,
-            storage_deposit_return_value: sdruc.value,
-            storage_deposit_return_inner_value: sdruc.inner,
+            storage_deposit_return_value: sdruc.value.to_string().parse().unwrap(),
+            storage_deposit_return_inner_value: sdruc.inner.to_string().parse().unwrap(),
```
Alex6323, line 117: I. see. unwraps.

src/analytics/ledger/active_addresses.rs
```diff
@@ -5,29 +5,34 @@ use std::collections::HashSet;
...
-    start_time: OffsetDateTime,
-    interval: Duration,
+#[allow(missing_docs)]
+pub struct AddressActivity {
+    pub start_time: OffsetDateTime,
+    pub interval: Duration,
     addresses: HashSet<Address>,
     // Unfortunately, I don't see another way of implementing it using our current trait design
     flush: Option<usize>,
 }
```
Alex6323, line 25: It would be nice if the `flush` field comment would also describe what the purpose of this field actually is (I know it's not part of this PR but maybe you can add it).

src/db/collections/analytics/base_token.rs
Alexandcoats: We don't do those conversions anymore. Those were because of mongodb.

src/db/collections/analytics/base_token.rs
Alexandcoats: These queries are going bye byte anyway

src/db/collections/analytics/base_token.rs
Alexandcoats: (Once the other PR is merged anyway)

src/analytics/ledger/ledger_outputs.rs
Alexandcoats: I am considering it., for sure.

src/db/collections/analytics/output_activity.rs
Alexandcoats: Yes. I think this is fixed in another PR though.

src/db/collections/analytics/milestone_size.rs
Alexandcoats: These are removed in another PR.

src/db/collections/analytics/unlock_condition.rs
Alexandcoats: Removed in another PR.

src/analytics/ledger/active_addresses.rs
Alexandcoats: I renamed it to `measurement` in another PR.

PR #1071, opened 2023-01-27
fix(db): slow newest/oldest milestone query

### LINKED ISSUE #1069: Fetching by `foundry_id` is very slow
The following query is slow even though the output is still _unspent_.

```
http://chronicle.testnet.shimmer.network/api/indexer/v1/outputs/foundry/0x082b0dc7b8c575e071d445a43d1a116aae4735891d73dea1f02e0aca74665ea0e90100000000
```

We should investigate why this is the case, especially because we should have an index on the `*_id` fields.

While we are at it, we should also make sure that our API also returns output ids when the `*_id` is _implicit_. From a quick glance it feels like we have a bug here too.

grtlr:
## Linked Issues

* Closes #1069.

## Notes to Reviewer

#### Database Changes

### DIFF
--- a/src/db/collections/milestone.rs
+++ b/src/db/collections/milestone.rs
@@ -247,7 +247,7 @@ impl MilestoneCollection {
         .await
     }
 
-    async fn get_milestone_by_order(&self, order: i32) -> Result<Option<MilestoneIndexTimestamp>, Error> {
+    async fn get_first_milestone_sorted(&self, order: i32) -> Result<Option<MilestoneIndexTimestamp>, Error> {
         self.aggregate(
             [
                 doc! { "$sort": doc! { "at.milestone_index": order } },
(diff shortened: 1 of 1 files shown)


Alex6323 (approved):

src/db/collections/milestone.rs
```diff
@@ -247,34 +247,31 @@ impl MilestoneCollection {
...
     }
 
-    /// Find the newest milestone.
-    pub async fn get_newest_milestone(&self) -> Result<Option<MilestoneIndexTimestamp>, Error> {
-        self.find_one(
-            doc! {},
-            FindOneOptions::builder()
-                .sort(doc! { "at.milestone_index": BY_NEWEST })
-                .projection(doc! {
+    async fn get_milestone_by_order(&self, order: i32) -> Result<Option<MilestoneIndexTimestamp>, Error> {
```
Alexandcoats, line 250: Maybe call this `get_milestone_sorted`?

grtlr: But it retrieves only the first element. So maybe `get_first_milestone_sorted`?

grtlr:
Please don't merge yet. Still need to check for the potential bug concerning alias ids in the indexer.

grtlr:
Yep there is a bug: #1074. But we should still merge this PR because those two bugs are unrelated.

PR #1075, opened 2023-01-27
fix(indexer)!: add indexed id to fix implicit aliases and nfts

### LINKED ISSUE #1074: Outputs with implicit alias and NFT ids are not indexed correctly
We currently don't handle the implicit ids of alias and NFT ids correctly.

We should create an additional field (with corresponding index) in `output.details` to help with the query.

Alexandcoats:
## Linked Issues

* Closes #1074

## Notes to Reviewer

#### API Changes
* [x] Test the indexer API endpoints.

#### INX Changes

#### Database Changes
* [x] Run the migration script.


Cargo.toml
```diff
@@ -129,3 +129,7 @@ inherits = "release"
 codegen-units = 1
 lto = true
 strip = "symbols"
+
+[[bin]]
+name = "1-0-0-beta-31"
+path = "src/migrations/1.0.0-beta.31.rs"
```
grtlr, line 135: We should add migrations as a subcommand because it makes it much more convenient to use in our released Docker image.

src/migrations/1.0.0-beta.31.rs
```diff
@@ -0,0 +1,109 @@
...
+    pub mongodb_database_name: String,
+}
+
+#[tokio::main]
+async fn main() -> eyre::Result<()> {
+    let args = MongoDbArgs::parse();
+    let config = MongoDbConfig {
+        conn_str: args.mongodb_conn_str,
+        database_name: args.mongodb_database_name,
+    };
```
grtlr, line 38: What's nice about subcommands is that they can reuse the already existing MongoDb args struct, so we could get rid of this code.

src/migrations/1.0.0-beta.31.rs
```diff
@@ -0,0 +1,109 @@
...
+        // Alias and nft are the same length so both can be done this way since they are just serialized as bytes
+        let id = AliasId::from(output_id);
+        collection
+            .update_one(
+                doc! { "$_id": output_id },
+                doc! { "$set": { "details.indexed_id": id } },
+                None,
+            )
+            .await?;
+    }
```
grtlr, line 77: Why not use a simple [`updateMany`](https://www.mongodb.com/docs/manual/reference/method/db.collection.updateMany/#mongodb-method-db.collection.updateMany) query?

There is no reason to retrieve all this data from the DB. It also makes the intent much clearer.

src/migrations/1.0.0-beta.31.rs
```diff
@@ -0,0 +1,109 @@
...
+        output_id: OutputId,
+    }
+
+    let outputs = collection
+        .aggregate::<Res>(
+            [
+                doc! { "$match": { "$or": [
+                    { "output.alias_id": AliasId::implicit() },
+                    { "output.nft_id": NftId::implicit() }
+                ] } },
```
grtlr, line 55: Aren't we missing the outputs with explicit ideas?

src/db/collections/outputs/indexer/mod.rs
```diff
@@ -75,16 +76,11 @@ impl OutputCollection {
...
-            IndexedId::Alias(_) => "output.alias_id",
-            IndexedId::Foundry(_) => "output.foundry_id",
-            IndexedId::Nft(_) => "output.nft_id",
-        };
         let mut res = self
             .aggregate(
                 [
                     doc! { "$match": {
-                        id_string: id,
+                        "output.details.indexed_id": id,
```
grtlr, line 83: I'm not sure this is correct. The indexer API differentiates between NFTs, aliases, and foundries. As far as I can tell we never check if the output returned from the database matches the queried output type.

Cargo.toml
grtlr: I don't think that's worth the updated complexity. The inflation of the binary is probably negligible.

Alex6323: We most definitely will have migrations in the future so a `migrate` subcommand makes sense in any case IMO. The only thing that this subcommand is then supposed to do is to check whether a certain migration is applicable by checking the versions and then execute the migration binary, or deny it with an appropriate message.

Alex6323: It's maybe negligible for the first migration :trollface:

src/migrations/1.0.0-beta.31.rs
```diff
@@ -0,0 +1,109 @@
...
+    )]
+    pub mongodb_conn_str: String,
+    /// The MongoDb database name.
+    #[arg(long, value_name = "NAME", default_value = mongocfg::DEFAULT_DATABASE_NAME)]
+    pub mongodb_database_name: String,
+}
+
+#[tokio::main]
+async fn main() -> eyre::Result<()> {
+    let args = MongoDbArgs::parse();
```
Alex6323, line 34: I think you forgot to read the environment variables.

src/migrations/1.0.0-beta.31.rs
Alex6323: yup, I agree. I think this `$match` should merely check for existence of those ids and in then make the distinction in the `for` loop below.

src/migrations/1.0.0-beta.31.rs
Alex6323: It's just a migration script. I would handle both cases in the same loop with `updateOne` tbh.

src/db/collections/outputs/indexer/mod.rs
Alex6323: Those ids are unique across NFTs, Aliases and Foundries ... so querying for a specific id will return the right output type (as long as we stored/computed the explicit id correctly), right?

Cargo.toml
Alexandcoats: The other option is to create/find an implementation of blake2b that we can use in a mongo script.

src/migrations/1.0.0-beta.31.rs
Alexandcoats: Again, `updateMany` cannot work that way. It can only update multiple items that all have the *same* (i.e. non-unique) key.

src/migrations/1.0.0-beta.31.rs
Alexandcoats: Ah, good point. Don't know what I was thinking.

src/db/collections/outputs/indexer/mod.rs
Alexandcoats: That, and also we check the `output.kind` in our queries anyway.

src/migrations/1.0.0-beta.31.rs
Alexandcoats: Good point, though it seems this will move anyway.

Cargo.toml
grtlr: @Alex6323 found a JavaScript one at some point. But I think we should just use our in-memory implementation.

src/migrations/1.0.0-beta.31.rs
Alexandcoats: Nevermind, I misread your initial comment.

src/db/collections/outputs/indexer/mod.rs
grtlr: I'm talking about the `indexed_output_by_id` query in particular. If a user requests an alias output by using `/api/indexer/v1/outputs/alias/{aliasId}` but for some reason provides an `foundryId`, the response should be either `400` or `404`.

Alex6323: you're right... I suggest adding:
```rust
impl IndexedId {
    fn kind(&self) -> &str {
        match self {
            Self::Alias(_) => AliasOutput::KIND,
            Self::Foundry(_) => FoundryOutput::KIND,
            Self::Nft(_) => NftOutput::KIND,
        }
    }
}
```
to further restrict the `$match`.

grtlr (approved):
THanks for adding the additional changes. LGTM, but still needs testing.

src/bin/inx-chronicle/migrations/migrate_1_0_0_beta_31.rs
```diff
@@ -0,0 +1,120 @@
...
+                ] } },
+                doc! { "$project": {
+                    "output_id": "$_id"
+                } },
+            ],
+            None,
+        )
+        .await?
+        .map_ok(|res| res.output_id)
+        .try_collect::<Vec<_>>()
```
Alex6323, line 38: We probably shouldn't collect those into memory but rather stream them, I guess.

src/bin/inx-chronicle/migrations/migrate_1_0_0_beta_31.rs
```diff
@@ -0,0 +1,120 @@
...
+
+use chronicle::{
+    db::{collections::OutputCollection, MongoDb, MongoDbCollection, MongoDbCollectionExt},
+    types::stardust::block::output::{AliasId, NftId, OutputId},
+};
+use futures::TryStreamExt;
+use mongodb::{bson::doc, options::IndexOptions, IndexModel};
+use serde::Deserialize;
+
+pub const PREV_VERSION: &str = "1.0.0-beta.30";
```
Alex6323, line 12: Why is this necessary? The version to run the migration from is already in the Cargo.toml manifest, right?

src/bin/inx-chronicle/migrations/migrate_1_0_0_beta_31.rs
grtlr: I agree that would be the correct way of doing things but I believe we will get rid of these migrations for a final RC anyways. So I would not worry about it now.

src/bin/inx-chronicle/migrations/migrate_1_0_0_beta_31.rs
Alexandcoats: Not if you've updated the application already.

src/bin/inx-chronicle/migrations/migrate_1_0_0_beta_31.rs
```diff
@@ -57,9 +57,9 @@ pub async fn migrate(db: &MongoDb) -> eyre::Result<()> {
                 "output.kind": "alias",
                 "output.alias_id": { "$ne": AliasId::implicit() },
             } },
-            doc! { "$set": {
+            vec![doc! { "$set": {
```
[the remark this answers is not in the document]
Alexandcoats: Unfortunately no 🤷 Apparently the update modification type does not `impl From<[T]>`.

grtlr (approved):

Alex6323 (approved):

grtlr (approved):

PR #1079, opened 2023-01-31
implement in-memory unlock condition analytics

grtlr:


src/analytics/ledger/unlock_conditions.rs
```diff
@@ -0,0 +1,139 @@
...
+#[allow(missing_docs)]
+pub struct UnlockConditionAnalytics {
+    pub timelock: CountValue,
+    pub expiration: CountValue,
+    pub storage_deposit_return: CountValue,
+    pub storage_deposit_return_inner_value: u64,
+}
+
+impl UnlockConditionAnalytics {
+    /// Initialize the analytics be reading the current ledger state.
```
Alexandcoats, line 23: *by

src/analytics/ledger/unlock_conditions.rs
```diff
@@ -0,0 +1,139 @@
...
+    pub fn init<'a>(unspent_outputs: impl IntoIterator<Item = &'a LedgerOutput>) -> Self {
+        let mut measurement = Self::default();
+        for output in unspent_outputs {
```
Alexandcoats, line 26: Since this is a lot of code, and we do the same thing below, maybe we should put it in a separate fn?

Alexandcoats: Actually, we could just use the init fn to build up an object then add/subtract it from self

src/analytics/ledger/unlock_conditions.rs
```diff
@@ -19,41 +19,45 @@ pub struct UnlockConditionAnalytics {
...
-                        measurement.timelock += output;
-                    }
-                    if basic.expiration_unlock_condition.is_some() {
-                        measurement.expiration += output;
-                    }
-                    if let Some(storage) = basic.storage_deposit_return_unlock_condition {
-                        measurement.storage_deposit_return += output;
-                        measurement.storage_deposit_return_inner_value += storage.amount.0;
-                    }
+fn add_unspent_outputs<'a>(measurement: &mut UnlockConditionAnalytics, unspent_outputs: impl IntoIterator<Item = &'a LedgerOutput>) {
```
Alexandcoats, line 22: Why isn't this a method on UnlockConditionAnalytics?

grtlr: Fair point

grtlr: I like free functions though

PR #1088, opened 2023-02-01
chore(ci): `check-all-features` on ready PRs

### LINKED ISSUE #1084: Investigate adding conditional CI
If it's possible to only run CI when a PR is reviewed, or a tag is added, or otherwise, we should try to do that for the more intensive CI checks like check features.

grtlr:
## Linked Issues

* Closes #1084.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,6 +13,11 @@ on:
       - "**.md"
       - "documentation/**"
       - "docker/**"
+    types:
+      - opened
+      - reopened
+      - synchronize
+      - ready_for_review
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -34,10 +39,12 @@ jobs:
     with: { os: ubuntu-latest, rust: nightly }
 
   clippy:
+    if: ${{ !github.event.pull_request.draft }}
```
Alex6323, line 42: Did you read through the discussion here? There seemed to be some way of bypassing the CI if one merged too quickly. Did you check whether that can still happen? https://github.com/orgs/community/discussions/25722

grtlr: From the discussion: https://github.com/orgs/community/discussions/25722#discussioncomment-3248917

This is exactly what we are doing, and it works nicely. I already tried it on this PR.

Alex6323: I mean what the guy writes right below: https://github.com/orgs/community/discussions/25722#discussioncomment-3248918

Alex6323 (approved):

.github/workflows/ci.yml
grtlr: I think this is an acceptable risk, as we also require at least one review.

PR #1091, opened 2023-02-01
`Vec`-based `LedgerUpdateStore`

### LINKED ISSUE #1082: Reimplement `LedgerUpdateStore` using `Vec` + `Map` to allow chunking
The LedgerUpdateStore cannot be chunked, so we should implemented it so that it can, and can also be keyed by OutputIds. It may also help to be able to get the outputs as a stream.

grtlr:
## Linked Issues

* Closes #1082.

### DIFF
--- a/src/bin/inx-chronicle/stardust_inx/error.rs
+++ b/src/bin/inx-chronicle/stardust_inx/error.rs
@@ -16,6 +16,15 @@ pub enum InxWorkerError {
     InfluxDb(#[from] influxdb::Error),
     #[error("expected INX address with format `http://<address>:<port>`, but found `{0}`")]
     InvalidAddress(String),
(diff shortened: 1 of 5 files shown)


src/bin/inx-chronicle/stardust_inx/error.rs
```diff
@@ -16,12 +16,6 @@ pub enum InxWorkerError {
     InfluxDb(#[from] influxdb::Error),
     #[error("expected INX address with format `http://<address>:<port>`, but found `{0}`")]
     InvalidAddress(String),
-    #[error("wrong number of ledger updates: `{received}` but expected `{expected}`")]
```
Alexandcoats, line 19: I think we should leave these until we move the error and make sure we don't need them.

PR #1095, opened 2023-02-01
Implement analytics using dyn Traits

Alexandcoats:
Tried to reduce the number of huge enums and match statements in our analytics, here is the result


src/analytics/ledger/active_addresses.rs
```diff
@@ -5,29 +5,26 @@ use std::collections::HashSet;
...
-use crate::types::{
-    ledger::{LedgerOutput, LedgerSpent, MilestoneIndexTimestamp},
-    stardust::block::Address,
-};
+use super::*;
+use crate::types::stardust::block::Address;
 
-pub struct AddressActivityMeasurement {
-    pub count: usize,
+pub(crate) struct AddressActivityMeasurement {
```
grtlr, line 11: Could we even use `pub(super)`?

Alexandcoats: It turns out we cannot, as `pub(super)` types cannot be re-exported.

PR #1096, opened 2023-02-02
Rename `OuputAmount` to `TokenAmount`

### LINKED ISSUE #1087: Consider usage of `OutputAmount` for analytics
We may want to use `OutputAmount` vs `u64` in analytics structs

grtlr:
## Linked Issues

* Closes #1087


Alexandcoats:
We should probably be consistent with either `amount` or `value`, right?

grtlr:
Should we do `value` -> `amount`?

Alex6323:
Lgtm!

PR #1097, opened 2023-02-02
fix(db)!: migration version checking

Alexandcoats:


src/bin/inx-chronicle/main.rs
```diff
@@ -44,6 +44,26 @@ async fn main() -> eyre::Result<()> {
...
+        .get_last_migration()
+        .await?
+        .as_deref()
+    {
+        None => {
+            db.collection::<ApplicationStateCollection>()
+                .set_last_migration(migrations::LATEST_VERSION)
+                .await?;
+        }
+        Some(migrations::LATEST_VERSION) => (),
```
grtlr, line 58: Maybe we should log that everything is up to date.

src/bin/inx-chronicle/migrations/mod.rs
```diff
@@ -1,22 +1,40 @@
...
 
-pub async fn migrate(version: &str, db: &MongoDb) -> eyre::Result<()> {
-    let curr_version = std::env!("CARGO_PKG_VERSION");
-    match version {
-        "1.0.0-beta.31" => {
-            if migrate_1_0_0_beta_31::PREV_VERSION == curr_version {
-                migrate_1_0_0_beta_31::migrate(db).await?;
-            } else {
-                bail!("cannot migrate to {} from {}", version, curr_version);
+const MIGRATIONS: [(Option<&str>, MigrateFn); 1] = [
```
grtlr, line 15:
```suggestion
const MIGRATIONS: &[(Option<&str>, MigrateFn)] = &[
```

src/bin/inx-chronicle/main.rs
Alexandcoats: I'd rather only log in the other cases, but if you feel strongly about it I can add it.

src/bin/inx-chronicle/migrations/migrate_20230202.rs
```diff
@@ -2,16 +2,23 @@
...
+        MongoDb, MongoDbCollection, MongoDbCollectionExt,
+    },
     types::stardust::block::output::{AliasId, NftId, OutputId},
 };
 use futures::TryStreamExt;
 use mongodb::{bson::doc, options::IndexOptions, IndexModel};
 use serde::Deserialize;
 
-pub const PREV_VERSION: &str = "1.0.0-beta.30";
+pub const VERSION: &str = "20230202";
```
grtlr, line 15:
```suggestion
pub const IDENTIFIER: &str = "20230202";
```

src/db/collections/application_state.rs
```diff
@@ -54,4 +58,25 @@ impl ApplicationStateCollection {
...
+            .await?
+            .and_then(|doc| doc.last_migration))
+    }
+
+    /// Set the current version in the singleton application state.
+    pub async fn set_last_migration(&self, last_migration: &str) -> Result<(), Error> {
+        self.update_one(
+            doc! {},
+            doc! {
+                "$set": { "last_migration": last_migration }
```
grtlr, line 75: I would write the current `PKG_VERSION` that was used to perform the migration into the struct too.

Alexandcoats: But no use-case

grtlr: It may help with debugging

grtlr: Feel free to resolve this though if you don't think it's worth it.

src/bin/inx-chronicle/migrations/mod.rs
```diff
@@ -1,22 +1,50 @@
...
+            .await?;
+        if matches!(last_migration.as_deref(), Some(LATEST_VERSION)) {
+            break;
+        }
+        match migrations.get(&last_migration.as_deref()) {
+            Some(migrate) => {
+                (migrate)(db)?;
+            }
+            None => {
+                bail!("cannot migrate version {:?}", last_migration);
```
grtlr, line 45: We should provide some standardized logging here and not rely on the migration providing the logging.

Alexandcoats: What do you mean?

src/bin/inx-chronicle/main.rs
```diff
@@ -44,6 +44,27 @@ async fn main() -> eyre::Result<()> {
...
+        None => {
+            info!("Setting migration version to {}", migrations::LATEST_VERSION);
+            db.collection::<ApplicationStateCollection>()
+                .set_last_migration(migrations::LATEST_VERSION)
+                .await?;
+        }
+        Some(migrations::LATEST_VERSION) => (),
+        Some(v) => {
+            eyre::bail!(
+                "Invalid database migration version {}, please run the `migrate` command.",
```
Alex6323, line 62: Small suggestion:
```rust
eyre::bail!("Expected database migration version {}, but found {}. Please run the `migrate` command.", migrations::LATEST_VERSION, v);
```

src/bin/inx-chronicle/migrations/migrate_20230202.rs
```diff
@@ -0,0 +1,122 @@
...
+use mongodb::{bson::doc, options::IndexOptions, IndexModel};
+use serde::Deserialize;
+
+use super::Migration;
+
+pub struct Migrate;
+
+#[async_trait]
+impl Migration for Migrate {
+    const VERSION: &'static str = "20230202";
```
Alex6323, line 19: How can a date be a version? Why can't we just use a simple version number that gets incremented with each migration?

src/db/collections/application_state.rs
```diff
@@ -15,7 +15,8 @@ use crate::{
 /// The MongoDb document representation of singleton Application State.
 #[derive(Clone, Debug, Eq, PartialEq, Serialize, Deserialize)]
 pub struct ApplicationStateDocument {
-    pub starting_index: MilestoneIndexTimestamp,
+    pub starting_index: Option<MilestoneIndexTimestamp>,
+    pub last_migration: Option<String>,
```
Alex6323, line 19: I think this should be a number instead of a string. Also would have the benefit that it's easy too determine how much versions the db is behind. Also the current design doesn't allow two migrations at the same day for now real reason.

src/bin/inx-chronicle/main.rs
```diff
@@ -44,6 +44,27 @@ async fn main() -> eyre::Result<()> {
...
+                v
+            );
+        }
+    }
```
Alex6323, line 66: Can we refactor this into a method (`handle_migration`) and inline it? Reason being that it takes up quite a few lines in our main function, but it shouldn't since it runs only in rare cases.

src/bin/inx-chronicle/migrations/mod.rs
```diff
@@ -1,22 +1,94 @@
...
+            .await?;
+        if matches!(last_migration.as_deref(), Some(LATEST_VERSION)) {
+            break;
+        }
+        match MIGRATIONS.get(&last_migration.as_deref()) {
+            Some(migration) => {
+                migration.migrate(db)?;
+            }
+            None => {
+                bail!("cannot migrate version {:?}", last_migration);
```
Alex6323, line 89: I think that means that the database got corrupted by the user somehow?! I mean ... that's fatal.

src/db/mongodb/collection.rs
```diff
@@ -65,6 +68,28 @@ pub trait MongoDbCollectionExt: MongoDbCollection {
...
+    async fn drop_index(
+        &self,
+        name: impl AsRef<str> + Send + Sync,
+        options: impl Into<Option<DropIndexOptions>> + Send + Sync,
+    ) -> Result<(), Error> {
+        match self.collection().drop_index(name, options).await {
+            Err(e) => match &*e.kind {
+                ErrorKind::Command(c) => {
+                    if c.code == INDEX_NOT_FOUND_CODE {
+                        Ok(())
```
Alex6323, line 82: really? :see_no_evil:

src/bin/inx-chronicle/main.rs
Alexandcoats: Well, there is no expected version, just a list of migrations and the versions they are from. And I dunno, I get that the command may change but I think it's better for errors to be explicit when telling you what to do.

src/bin/inx-chronicle/migrations/migrate_20230202.rs
Alexandcoats: Technically it's not a date, it's a string. Serious answer: I wanted something that could could be easily understood and would tell you when it happened. As Jochen pointed out, we could still do two (or more) migrations in one day, but let's be honest, that will never happen.

src/db/collections/application_state.rs
```diff
@@ -54,4 +58,25 @@ impl ApplicationStateCollection {
         .await?;
         Ok(())
     }
+
+    /// Gets the singleton application state.
```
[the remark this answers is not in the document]
Alexandcoats: oop

src/bin/inx-chronicle/migrations/mod.rs
Alexandcoats: Well, yeah. What else should I say though? "You monkeyed with the database, why did you do that? Please don't contact support."

src/bin/inx-chronicle/main.rs
Alex6323: Aren't we expecting the db to have been migrated to the LATEST_VERSION?

grtlr: Let's get this merged, we can always improve the logging.

src/bin/inx-chronicle/main.rs
grtlr: I would love for stuff to be as self-contained as possible, so I would like that. But we can open an issue for it if you don't want to do that now.

Alex6323: You're right that is subjective like many things in the end, but for those "controller" like functions (call A, then B, then C, then D, ...) it's nice to refactor the logic out into small, self-describing sub-routines that can (and probably will automatically) be inlined. By picking good names one gets a quick idea what `main` is actually doing without the reader having to parse undocumented code.

src/db/mongodb/collection.rs
grtlr: Thinking about this we don't need that code at all. The migration should fail if the indexes are not their, we could see it as a safeguard.

grtlr: If we ever want to ignore these errors it should be in the migration specific code.

src/bin/inx-chronicle/migrations/mod.rs
Alex6323: yeah, something like that, or ... I suggest printing something a little bit more helpful here that doesn't leave the user with question marks above their head, like: "cannot migrate from a corrupted database".

src/db/mongodb/collection.rs
Alex6323: alrighty ... let's resolve and never look at it again then :trollface:

src/bin/inx-chronicle/main.rs
```diff
@@ -44,6 +45,8 @@ async fn main() -> eyre::Result<()> {
         ByteSize::b(db.size().await?)
     );
 
+    check_migration_version(&db).await?;
+
     build_indexes(&db).await?;
```
grtlr, line 50: Ideally this would be a migration as well.

Alex6323 (dismissed):

Alex6323 (approved):

PR #1098, opened 2023-02-03
Minor cleanup of migration logic

grtlr:


src/bin/inx-chronicle/migrations/mod.rs
```diff
@@ -9,30 +9,21 @@ use eyre::bail;
...
+/// in the form `migrate_YYYYMMDD`.
+
 pub const LATEST_VERSION: &str = migrate_20230202::Migrate::VERSION;
 
-macro_rules! migration {
-    ($($mig:expr),*$(,)?) => {
-        migration([$(Box::new($mig) as _),*])
-    };
-}
+const MIGRATIONS: &[&'static dyn DynMigration] = &[&migrate_20230202::Migrate];
```
Alexandcoats, line 19: Only one minor comment: Can you put the entry into a newline so that it's more obvious that it's intended to be added to? You'll need to add a comment above the item to avoid formatting.

Alexandcoats (approved):

PR #1099, opened 2023-02-03
Fix in-memory analytics

grtlr:


src/db/collections/block.rs
```diff
@@ -264,23 +265,35 @@ impl BlockCollection {
...
-                     metadata,
-                 }| (block_id, block, raw, metadata),
-            ))
+            .map_ok(|r| {
+                (
+                    r.block_id,
+                    RawMessage::<iota_types::block::Block>::from(r.raw.clone())
+                        .inner_unverified()
+                        .unwrap()
+                        .into(),
```
Alex6323, line 292: Can you help me understand this better? So ... eventually we need the protocol parameters from the time where this `Block` was serialized and written to the db?

src/db/collections/block.rs
```diff
@@ -264,23 +265,35 @@ impl BlockCollection {
...
         index: MilestoneIndex,
     ) -> Result<impl Stream<Item = Result<(BlockId, Block, Vec<u8>, BlockMetadata), Error>>, Error> {
+        #[derive(Debug, Deserialize)]
+        struct QueryRes {
+            #[serde(rename = "_id")]
+            block_id: BlockId,
+            #[serde(with = "serde_bytes")]
+            raw: Vec<u8>,
+            metadata: BlockMetadata,
+        }
```
Alex6323, line 275: Why is this type necessary? Could we not just ignore `block` when mapping to our returned tuple? Or is this already the problem that the deserialization will fail?

src/db/collections/block.rs
grtlr: No, the problem is we don't store `Output`s in the `BlockCollection`

src/db/collections/block.rs
grtlr: Deserialization will fail. We should consider removing `derive(Deserialize)`.

Alexandcoats: It's unfortunate but we are allowed to gather the outputs by lookup, so it still needs to be `Deserialize`.

Alexandcoats (approved):

src/db/collections/block.rs
Alex6323: ah, I see, thanks

Alex6323 (approved):

src/db/mongodb/mod.rs
```diff
@@ -23,7 +23,8 @@ const DUPLICATE_KEY_CODE: i32 = 11000;
 /// A handle to the underlying `MongoDB` database.
 #[derive(Clone, Debug)]
 pub struct MongoDb {
-    pub(crate) db: mongodb::Database,
+    pub(crate) database_name: String,
+    //pub(crate) db: mongodb::Database,
```
Alexandcoats, line 27: fmt

PR #1110, opened 2023-02-06
fix(inx): node config updates

### LINKED ISSUE #1108: Write past node config into MongoDB
Now that we listen to confirmed milestones, we should also push the historical protocol parameters to the database. Right now we only push protocol parameters that are equal to or newer than the starting of the sync process.

Alexandcoats:
## Linked Issues

* Closes #1108 

## Notes to Reviewer

#### INX Changes
* [x] Ensure that node configs older than the `starting_index` are still written (Alex6323: I think that's covered by the tests).

#### Database Changes

### DIFF
--- a/src/db/collections/configuration_update.rs
+++ b/src/db/collections/configuration_update.rs
@@ -68,15 +68,11 @@ impl ConfigurationUpdateCollection {
         ledger_index: MilestoneIndex,
         config: NodeConfiguration,
     ) -> Result<(), Error> {
-        let params = self.get_node_configuration_for_ledger_index(ledger_index).await?;
-        if params.is_none()
-            || params
-                .map(|latest_config| latest_config.config != config)
-                .unwrap_or_default()
-        {
+        let node_config = self.get_node_configuration_for_ledger_index(ledger_index).await?;
+        if !matches!(node_config, Some(node_config) if node_config.config == config) {
             self.update_one(
                 doc! { "_id": ledger_index },
-                doc! { "$set": mongodb::bson::to_bson(&config).unwrap() },
+                doc! { "$set": mongodb::bson::to_bson(&config)? },
                 UpdateOptions::builder().upsert(true).build(),
             )
             .await?;
--- a/src/db/collections/configuration_update.rs
+++ b/src/db/collections/configuration_update.rs
@@ -1,7 +1,11 @@
 // Copyright 2022 IOTA Stiftung
 // SPDX-License-Identifier: Apache-2.0
 
-use mongodb::{bson::doc, error::Error, options::FindOneOptions};
+use mongodb::{
+    bson::doc,
+    error::Error,
+    options::{FindOneOptions, UpdateOptions},
+};
 use serde::{Deserialize, Serialize};
 
 use crate::{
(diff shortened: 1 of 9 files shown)


tests/node_configuration.rs
```diff
@@ -93,19 +93,6 @@ mod test_rand {
...
-            .await
-            .unwrap();
-        assert_eq!(node_configuration.count().await.unwrap(), 1);
-        assert_eq!(&latest_config.config.base_token.name, "Shimmer");
-        assert_eq!(&latest_config.config.base_token.ticker_symbol, "SMR");
-        assert_eq!(&latest_config.config.base_token.unit, "SMR");
-        assert_eq!(&latest_config.config.base_token.subunit, "glow");
-        assert_eq!(latest_config.config.base_token.decimals, 6);
-        assert!(latest_config.config.base_token.use_metric_prefix);
-
```
grtlr, line 108: I don't think we should delete this test case. We should rather "sandwich" a meaningless configuration _t1_ (where nothing has changed) between two configurations at _t0_ and _t2_. Ideally, _t0_ would be inserted after _t2_ to mimic our logic for catching up.

src/db/collections/protocol_update.rs
```diff
@@ -92,17 +92,17 @@ impl ProtocolUpdateCollection {
     }
 
     /// Add the protocol parameters to the list if the protocol parameters have changed.
-    pub async fn update_latest_protocol_parameters(
+    pub async fn upsert_protocol_parameters(
```
Alex6323, line 95: Should `insert_protocol_parameters` become private now as both cases (insertion, updating) are handled by this method, and the user can't keep track of whether it was inserted already anyway?

src/db/collections/configuration_update.rs
```diff
@@ -64,40 +68,18 @@ impl ConfigurationUpdateCollection {
...
-            if latest_config.ledger_index >= ledger_index {
-                return Ok(());
-            } else if latest_config.config != config {
-                self.insert_one(ConfigurationUpdateDocument { ledger_index, config }, None)
-                    .await?;
-            }
-        } else {
-            self.insert_one(ConfigurationUpdateDocument { ledger_index, config }, None)
-                .await?;
+        let params = self.get_node_configuration_for_ledger_index(ledger_index).await?;
```
Alex6323, line 71: `params` ~> `node_config`
`latest_config` ~> `node_config`

src/db/collections/configuration_update.rs
```diff
@@ -64,40 +68,18 @@ impl ConfigurationUpdateCollection {
...
+        if params.is_none()
+            || params
+                .map(|latest_config| latest_config.config != config)
+                .unwrap_or_default()
```
Alex6323, line 75: `if !matches!(node_config, Some(node_config) if node_config == config) { .. }`

src/db/collections/configuration_update.rs
```diff
@@ -64,40 +68,18 @@ impl ConfigurationUpdateCollection {
...
+        {
+            self.update_one(
+                doc! { "_id": ledger_index },
+                doc! { "$set": mongodb::bson::to_bson(&config).unwrap() },
```
Alex6323, line 79: Undocumented `unwrap`.

src/db/collections/protocol_update.rs
```diff
@@ -73,36 +77,26 @@ impl ProtocolUpdateCollection {
...
+                .unwrap_or_default()
+        {
+            self.update_one(
+                doc! { "_id": ledger_index },
+                doc! { "$set": {
+                    "parameters": mongodb::bson::to_bson(&parameters).unwrap()
+                } },
+                UpdateOptions::builder().upsert(true).build(),
+            )
+            .await?;
```
Alex6323, line 99: here as well ... I think `!matches!` is still preferable ... and the undocumented `unwrap`

src/db/collections/configuration_update.rs
Alexandcoats: This isn't exactly the first one for `to_bson`.

Alex6323 (approved):

PR #1130, opened 2023-02-14
Small refactor

Alex6323:


docker/assets/grafana/dashboards/analytics_dashboard.json
```diff
@@ -3209,7 +3191,7 @@
           "tags": []
         }
       ],
-      "title": "Consumed Output Distribution",
+      "title": "Number of Transactions per Number of Consumed Outputs /${aggregation_interval}",
```
Alexandcoats, line 3194: Overly verbose. Prefer the original title.

Alex6323: Fair enough, but I rather have the title verbose than the legend and the popup overlay. This title makes it clear as well what the numbers stand for while the original proposal had yo make this clear individually, which resulted in more verbosity than this solution. Also ... I wouldnt have picked it myself, but we have a lot of "Number of ..." in the other panels so I wanted it to be consistent.

Alexandcoats: I don't think it makes it clear at all.

Alex6323: what's unclear?

Alexandcoats: It simply does not describe what the stat is.

Alexandcoats (approved):

PR #1131, opened 2023-02-14
fix(analytics): fill daily analytics

Alexandcoats:
This PR adds the `fill-daily-analytics` command to the CLI and removes daily calculations from the regular workflow, including during syncing. This means the only way to gather daily analytics is by calling the CLI command.

This PR adds two additional indexes to support querying for outputs by timestamp.

## Reasoning
There are two main issues with using per-milestone data to calculate daily analytics.
1. The date transitions are not simple. We can only know that we should calculate a date by having two milestones and comparing the dates. This creates situations where it is possible to miss analytics even when syncing. Moreover, the code to support these calculations is over-complex.
2. Filling analytics per-milestone using tasks can miss date transitions, as one task can finish on the last milestone of a day, while another can begin on the next.

This also neatly removes the problem of running `fill-analytics` only for daily analytics (which would not be efficient).

INX data does not support this workflow, as the Unit of Work is a milestone, not a date. Instead, this PR uses MongoDB exclusively to gather these analytics. This is performant enough given that the query must only be run once per day.

## Future Work
It may be that we eventually want to run these analytics during syncing. Comparing milestone timestamps in this case may be "good enough" to support this.

### DIFF
--- a/src/bin/inx-chronicle/cli/mod.rs
+++ b/src/bin/inx-chronicle/cli/mod.rs
@@ -319,9 +319,10 @@ impl ClArgs {
                     return Ok(PostCommand::Exit);
                 }
                 #[cfg(feature = "analytics")]
-                Subcommands::FillDailyAnalytics {
+                Subcommands::FillIntervalAnalytics {
                     start_date,
                     end_date,
+                    interval,
                     num_tasks,
                     analytics,
                 } => {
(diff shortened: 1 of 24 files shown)


Alex6323:
I agree with the reasoning. It's unfortunate having to separate the CLI commands for this but still better than having wrong analytics.

src/bin/inx-chronicle/cli/mod.rs
```diff
@@ -373,6 +417,25 @@ pub enum Subcommands {
...
+    /// Fill interval analytics from Chronicle's database.
+    #[cfg(feature = "analytics")]
+    FillIntervalAnalytics {
+        /// The inclusive starting date (YYYY-MM-DD).
+        #[arg(short, long, value_parser = parse_date)]
+        start_date: Option<time::Date>,
+        /// The inclusive ending date (YYYY-MM-DD).
+        #[arg(short, long, value_parser = parse_date)]
+        end_date: Option<time::Date>,
+        /// The interval to use [day, week, month, year].
```
Alex6323, line 429: The possible values are rendered automatically. How nice of `clap`.
```suggestion
        /// The interval to use.
```

PR #1133, opened 2023-02-14
fix(api): refactor block queries to avoid lookups

### LINKED ISSUE #1105: API endpoints which request blocks will not return conflicting outputs
### Bug description

API endpoints like `core/v2/block`, if not requesting the raw block, will return blocks without the outputs if the transaction is conflicting.

### Chronicle version

- Chronicle version (version number, commit, or branch): 1.0.0-beta.30

Alexandcoats:
## Linked Issues

* Closes #1105 

## Notes to Reviewer

#### API Changes
* [x] Ensure that the API response times scale with database size appropriately.
* [x] Check for breaking changes in the API and matching (conventional) commit message prefix.

#### Database Changes

### DIFF
--- a/src/bin/inx-chronicle/api/stardust/poi/routes.rs
+++ b/src/bin/inx-chronicle/api/stardust/poi/routes.rs
@@ -179,7 +179,7 @@ async fn create_proof_for_applied_blocks(
 
     // Fetch the referenced and applied block ids in "White Flag" order, and make sure they contain the block.
     let applied_block_ids = block_collection
-        .get_applied_blocks_in_white_flag_order(referenced_index)
+        .get_referenced_blocks_in_white_flag_order(referenced_index)
         .await?;
     if !applied_block_ids.contains(&block_id) {
         return Err(RequestError::PoI(poi::RequestError::BlockNotApplied(block_id.to_hex())).into());
--- a/src/bin/inx-chronicle/api/stardust/poi/routes.rs
+++ b/src/bin/inx-chronicle/api/stardust/poi/routes.rs
@@ -179,7 +179,7 @@ async fn create_proof_for_applied_blocks(
 
     // Fetch the referenced and applied block ids in "White Flag" order, and make sure they contain the block.
     let applied_block_ids = block_collection
-        .get_referenced_blocks_in_white_flag_order(referenced_index)
+        .get_applied_blocks_in_white_flag_order(referenced_index)
         .await?;
     if !applied_block_ids.contains(&block_id) {
         return Err(RequestError::PoI(poi::RequestError::BlockNotApplied(block_id.to_hex())).into());
--- a/src/db/collections/block.rs
+++ b/src/db/collections/block.rs
@@ -268,6 +268,28 @@ impl BlockCollection {
             }))
     }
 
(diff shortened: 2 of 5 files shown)


src/bin/inx-chronicle/api/stardust/poi/routes.rs
```diff
@@ -179,7 +179,7 @@ async fn create_proof_for_applied_blocks(
 
     // Fetch the referenced and applied block ids in "White Flag" order, and make sure they contain the block.
     let applied_block_ids = block_collection
-        .get_applied_blocks_in_white_flag_order(referenced_index)
```
Alex6323, line 182: How is that not breaking PoI for applied transactions? :thinking:

src/db/collections/block.rs
```diff
@@ -132,64 +131,47 @@ pub struct IncludedBlockResult {
 
 #[derive(Deserialize, Debug, Clone)]
 pub struct IncludedBlockMetadataResult {
+    #[serde(rename = "_id")]
```
Alex6323, line 134: What about `IncludedBlockResult`? Should we rename there as well for consistency?

src/db/collections/block.rs
```diff
@@ -132,64 +131,47 @@ pub struct IncludedBlockResult {
...
-            ],
-            None,
-        )
-        .await?
-        .try_next()
-        .await
+        Ok(self
+            .get_block_raw(block_id)
+            .await?
+            .map(|raw| iota_types::block::Block::unpack_unverified(raw).unwrap().into()))
```
Alex6323, line 158: I have a very bad feeling about this line.

Alexandcoats: How so?

src/bin/inx-chronicle/api/stardust/poi/routes.rs
Alexandcoats: The problem is the `Included` check? I think I was assuming referenced blocks are never conflicting.

Alex6323: Yeah...that's the problem. Milestones can reference conflicting transactions.

src/db/collections/block.rs
Alex6323: Well ... what happens if the binary representation changes in the future? How do we then unpack old Blocks if we do it unverified?

src/bin/inx-chronicle/api/stardust/poi/routes.rs
Alexandcoats: Makes sense, my mistake. I'll revert this change.

src/db/collections/block.rs
Alexandcoats: If the representation changes, there will have to be a version change in `iota_types`, and most likely there will have to be a solution in that crate for old representations.

Alex6323: I am unsure here. Arent' the protocol parameters all it needs as context to unpack older versions of blocks?

Alex6323: right, thx

Alex6323 (approved):

PR #1138, opened 2023-02-17
Set up buckets only once for Transaction Size analytic

Alex6323:


Alexandcoats:
Why is this based on the `fill-daily` branch? Also, it would probably be cleaner to create a static `[(Bucket, usize); 11]` and just convert it to a hashmap each milestone.

Alex6323:
Less work I suppose updating this file again when `fill-daily` gets merged.

PR #1171, opened 2023-03-07
Decode hex string

### LINKED ISSUE #1165: Chronicle doesn't return outputs compared to Hornet
### Bug description

There is a case found where the response from Chronicle is different from the Hornet response.

Endpoint: `/api/indexer/v1/outputs/basic`

Usage with `tag` query param.

Example:
tag: "Soonaverse" (0x536f6f6e617665727365)
Chronicle -> https://chronicle.testnet.shimmer.network/api/indexer/v1/outputs/basic?tag=0x536f6f6e617665727365
VS
Hornet -> https://api.testnet.shimmer.network/api/indexer/v1/outputs/basic?tag=0x536f6f6e617665727365

### Steps to reproduce the bug

1. View and compare the responses of links above

### Expected behaviour

Chronicle should return the outputs. Response should be identical as when using Hornet for same input.

### Actual behaviour

Chronicle return an empty array.

Alex6323:
## Linked Issues

* Closes #1165 

## Notes to Reviewer

#### API Changes

#### Database Changes

### DIFF
--- a/src/db/mongodb/collections/outputs/indexer/queries.rs
+++ b/src/db/mongodb/collections/outputs/indexer/queries.rs
@@ -66,11 +66,13 @@ pub(super) struct TagQuery(pub(super) Option<String>);
 impl AppendToQuery for TagQuery {
     fn append_to(self, queries: &mut Vec<Document>) {
         if let Some(tag) = self.0 {
+            // `tag` is a hex string, so we need to decode it first.
+            let tag = prefix_hex::decode::<Vec<u8>>(&tag).unwrap();
             queries.push(doc! {
                 "output.features": {
                     "$elemMatch": {
                         "kind": "tag",
-                        "data": bson::to_bson(&serde_bytes::Bytes::new(tag.as_bytes())).unwrap()
+                        "data": bson::to_bson(&serde_bytes::ByteBuf::from(tag)).unwrap()
                     }
                 }
             });


src/model/block/payload/transaction/output/mod.rs
```diff
@@ -261,6 +263,46 @@ impl From<Output> for iota::dto::OutputDto {
...
+        Ok(Self(prefix_hex::decode::<Vec<u8>>(tag.as_ref())?))
+    }
+
+    /// Converts the [`Tag`] to its `0x`-prefixed hex representation.
+    pub fn to_hex(&self) -> String {
+        prefix_hex::encode(&*self.0)
+    }
+}
+
+// Note: assumes an ASCII string as input.
```
Alexandcoats, line 283: This should say utf-8, not ascii, right?

Alex6323: I am not sure ... `String`s obviously support `utf-8`, but I think IOTA assumes tags to be ASCII.

Alexandcoats: Well the fact is rust strings are not ASCII, so this comment is odd. Since we only use it for testing, I think we should just remove the comment.

Alexandcoats (approved):

PR #1176, opened 2023-03-08
fix(inx): remove startup checks and revert influx update that breaks analytics

Alexandcoats:
# Description

This PR removes the unnecessary checks at startup that verify the protocol parameters. This behavior was also causing an issue where a protocol update would break Chronicle if the network name changed and Chronicle was restarted before it reached the milestone with the new params.

## Notes to Reviewer

#### INX Changes

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1279,9 +1279,9 @@ dependencies = [
 
 [[package]]
 name = "influxdb"
-version = "0.6.0"
+version = "0.5.2"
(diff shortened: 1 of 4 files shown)


Alex6323:
I think we shouldn't remove the protocol parameter's check, because it's conceivable that a user accidentally connects Chronicle to a different node running in a different network, and then this would corrupt the database. Most users won't have the knowledge to fix the database after this, and switching to a new network name while keeping the previous data is not a standart/normal use-case.

PR #1186, opened 2023-03-13
feat(cli)!: consolidate fill analytics once again

### LINKED ISSUE #1175: Allow CLI `fill-analytics` to be run using dates
## Description

Currently `fill-analytics` only accepts milestones as the range. It would be very helpful for some situations to alternatively accept dates.

Alexandcoats:
This PR re-merges `fill-analytics` and `fill-interval-analytics` while also allowing the `--start-date` and `--end-date` flags to determine the start and end milestones for per-milestone analytics. The rules for this are outlined in the CLI docs:

```
This commands accepts both milestone index and date ranges. The following rules apply:
- If both milestone and date are specified, the date will be used for interval analytics
while the milestone will be used for per-milestone analytics.
- If only the milestone is specified, the date will be inferred from the milestone timestamp.
- If only the date is specified, the milestone will be inferred from the available data from that date.
- If neither are specified, then the entire range of available data will be used.
```

I wish I had thought to do things this way before we released the analytics changes, but better late than never I guess 🤷‍♀️

## Linked Issues

* Closes #1175

## Notes to Reviewer

#### CLI Changes
* [ ] Run the new `fill-analytics` using various combinations of flags.

### DIFF
--- a/src/bin/inx-chronicle/cli/analytics.rs
+++ b/src/bin/inx-chronicle/cli/analytics.rs
@@ -23,7 +23,7 @@ use tracing::{debug, info};
 
 use crate::config::ChronicleConfig;
 
-/// This commands accepts both milestone index and date ranges. The following rules apply:
+/// This command accepts both milestone index and date ranges. The following rules apply:
 ///
 /// - If both milestone and date are specified, the date will be used for interval analytics
 /// while the milestone will be used for per-milestone analytics.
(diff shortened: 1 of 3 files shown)


src/bin/inx-chronicle/cli/analytics.rs
```diff
@@ -21,88 +23,178 @@ use tracing::{debug, info};
 
 use crate::config::ChronicleConfig;
 
+/// This commands accepts both milestone index and date ranges. The following rules apply:
```
Alex6323, line 26:
```suggestion
/// This command accepts both milestone index and date ranges. The following rules apply:
```

Alex6323 (approved):

PR #1187, opened 2023-03-14
chore(cleanup): Improve analytics measurement org

Alex6323:


Alexandcoats:
I don't like having influx-specific logic in our analytics, which are general purpose. If we ever update to influx 2 or use a different time series we would then need to change all of these files rather than just one place.

Alex6323:
Sure, but that's a hypothetical reason. I look at it from a reader perspective. If I want a good picture over our analytic types I want as much information as possible together in one file. This simplifies debugging or changes because I one change in that type only changes a single file rather than 2, and this is much more likely situation than the one that you described.

Alexandcoats:
Both are hypothetical reasons. I see your point but changing two files is much less burdensome than changing 16, in the worst cases.

Alex6323:
honestly, I am not worried about file changes. This PR didn't take long despite having to change all those. I am much more worried about stupid bugs like forgetting adding a new field to the influx query.

Alexandcoats:
The IDE isn't going to write a new time series implementation for you. Also, forgetting to add a field can easily happen even if the implementation is in the same file. The only way to guarantee we wouldn't forget is to use a property of rust which would not compile otherwise, and that would happen regardless of the locations.

Alex6323:
No, that's the only way we wouldn't forget it 100% of the time. I just want everything in one place because it's less likely to make such mistakes, not because it gives any guarantees.

Alexandcoats:
Okay, but let's say we implemented a way for the compiler to ensure we use all the fields. Then, does it matter if they are in the same file anymore? Wouldn't it in fact be better for them to be contained together for the reason I mentioned?

Alex6323:
didn't make it :cry:

PR #1188, opened 2023-03-14
chore(grafana): Don't interpolate before 1st datapoint in series

Alex6323:


Alexandcoats:
So strange. I still have never encountered this on my instance.

Alexandcoats:
So I found the [docs](https://grafana.com/docs/grafana/latest/panels-visualizations/visualizations/time-series/#line-interpolation) about that parameter. I don't think it has any relationship with the bug you are encountering.

PR #1191, opened 2023-03-14
fix(db): add parent-child table to avoid big index

### LINKED ISSUE #1190: `block_parents_index` is too big
### Bug description

On the shimmer beta chronicle instance, this index is almost 8GB. This is much too large. We must find a way to provide the benefits of an index without trying to keep all of that in memory. Perhaps a materialized view?

### Chronicle version

- Chronicle version: `1.0.0-beta.36`

Alexandcoats:
## Linked Issues

* Closes #1190

## Notes to Reviewer

#### API Changes
* [ ] Review the API documentation changes and confirm that it matches the actual functionality.
* [ ] Check for breaking changes in the API and matching (conventional) commit message prefix.

#### Database Changes

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -364,7 +364,7 @@ checksum = "baf1de4339761588bc0619e3cbc0120ee582ebb74b53b4efbf79117bd2da40fd"
 
 [[package]]
 name = "chronicle"
-version = "1.0.0-beta.36"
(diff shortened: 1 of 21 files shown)


Alexandcoats:
Unfortunately this solution is not better than the current one 🤷‍♀️

PR #1199, opened 2023-03-20
fix(logging): cannot migrate error message

### LINKED ISSUE #1198: `Option` (implementation detail) leaks into log output
### Bug description

I don't like that the `Option` leaks into the log ... we should call `.unwrap_or(0)` on it.

image

Alex6323:
## Linked Issues

* Closes #1198

### DIFF
--- a/src/bin/inx-chronicle/migrations/mod.rs
+++ b/src/bin/inx-chronicle/migrations/mod.rs
@@ -120,7 +120,7 @@ pub async fn migrate(db: &MongoDb) -> eyre::Result<()> {
             None => {
                 bail!(
                     "cannot migrate version {:?}, database is in invalid state",
-                    last_migration
+                    last_migration.unwrap_or(0)
                 );
             }
         }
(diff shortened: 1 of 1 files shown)


Alexandcoats (changes requested):
Nope, this cannot work. Migration 0 is a valid version, so printing this would be wrong. I personally don't see the issue with printing an option value (we do it in many places, I am certain), but if you really want to remove that then you'll have to print two different messages based on the value. Or you could just unwrap it, since `None` should always lead to a valid migration.

Alex6323:
yeah, I got confused and took the `None` for version `0`, but it's actually `Some(0)` what we're getting, so this cannot work.

src/bin/inx-chronicle/migrations/mod.rs
```diff
@@ -118,9 +118,10 @@ pub async fn migrate(db: &MongoDb) -> eyre::Result<()> {
                 migration.migrate(db).await?;
             }
             None => {
+                // Unwrap: it's safe to unwrap `last_migration` in this case.
```
Alexandcoats, line 121: Perhaps explain why?

Alex6323: sure

Alexandcoats (approved):

PR #1207, opened 2023-03-21
chore(cleanup): spring cleaning

Alex6323:


src/bin/inx-chronicle/main.rs
```diff
@@ -107,17 +105,18 @@ async fn main() -> eyre::Result<()> {
...
             worker.run(handle.recv().then(|_| async {})).await?;
             Ok(())
         });
     }
 
     // We wait for either the interrupt signal to arrive or for a component of our system to signal a shutdown.
     tokio::select! {
-        _ = process::interrupt_or_terminate() => {
+        res = process::interrupt_or_terminate() => {
+            res?;
```
Alexandcoats, line 119: Let's not do this

Alex6323: Don't let me always ask why and come forward with an actual explanation please other than I don't like. You know I can be reasoned with, but I would never give you a comment like this without also explaining what I don't like about it.

src/bin/inx-chronicle/process.rs
```diff
@@ -1,28 +1,19 @@
...
-pub async fn interrupt_or_terminate() {
+pub async fn interrupt_or_terminate() -> eyre::Result<()> {
     #[cfg(unix)]
     {
         use tokio::signal::unix::{signal, SignalKind};
-
-        let mut sigterm = signal(SignalKind::terminate()).expect("cannot listen to `SIGTERM`");
-        let mut sigint = signal(SignalKind::interrupt()).expect("cannot listen to `SIGINT`");
-
+        let mut terminate = signal(SignalKind::terminate())?;
```
Alexandcoats, line 8: Rather than remove these strings, can we replace them with `eyre` wrappers?

```
let mut terminate = signal(SignalKind::terminate()).map_err(|e| eyre!("cannot listen to `SIGTERM`: {e}"));
```

src/bin/inx-chronicle/process.rs
```diff
@@ -1,28 +1,19 @@
...
+        let mut interrupt = signal(SignalKind::interrupt())?;
         tokio::select! {
-            _ = sigterm.recv() => {
-                tracing::info!("received `SIGTERM`, sending shutdown signal")
-            }
-            _ = sigint.recv() => {
-                tracing::info!("received `SIGINT`, sending shutdown signal")
-            }
+            _ = terminate.recv() => {}
```
Alexandcoats, line 11: Please don't remove these logs.

Alex6323: we already log this in the caller. It's also wrong, because how does this function know what's going on outside of this function.

src/bin/inx-chronicle/main.rs
```diff
@@ -131,7 +130,8 @@ async fn main() -> eyre::Result<()> {
 
     // Allow the user to abort if the tasks aren't shutting down quickly.
     tokio::select! {
-        _ = process::interrupt_or_terminate() => {
+        res = process::interrupt_or_terminate() => {
+            res?;
```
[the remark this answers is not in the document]
Alex6323: Well, that's a good point I can absolutely agree with. In my defense tho ... that is a functional change in behavior, because before ... we just panicked.

Alexandcoats (approved):

PR #1211, opened 2023-03-23
Update docker-compose.yml

ahratul:
I added "--p2p.autopeering.enabled=true” to this line, otherwise, the hornet will not be auto peering and it will not work.

Ways to reproduce:
* Run `docker compose` as-is without modifying it.

(edited by @Alex6323 : I allowed me to change the description to remove the redundant information)

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -84,6 +84,7 @@ services:
       # - "-c"
       # - "config_testnet.json"
       # - "config_alphanet.json"
+      - "--p2p.autopeering.enabled=true”
       - "--inx.enabled=true"
       - "--inx.bindAddress=hornet:9029"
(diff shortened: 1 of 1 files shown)


Alex6323:
Thanks for the PR. But we cannot and should not enable autopeering by default. I guess we should mention this in the docs, that you first have to add this line to the `docker-compose.yml` file. We can however, already put it in there but leave it out-commented, i.e. disabled by default, and mention to uncomment/include it if autopeering is wanted and the security implications of enabling it are accepted by the user.

ahratul:
Hi, Sorry for the late reply

Let's mention them in the yml file. Can we comment on the yml file? That will help if someone new wants to start work (like me) it will help.

PR #1230, opened 2023-05-05
fix: forward worker errors to main exit code

### LINKED ISSUE #1228: INX-chronicle exists with wrong exit code and thus does not restart
### Bug description
When chronicle times out with influx it shuts down successfully and is not being restarted by docker.image

Chronicle should exit with a different (failure error) code so that docker restarts it when there is a hiccup with influx. It should not report as "successful shutdown" in that case.

### Rust version

- Rust version:

### Chronicle version

- Chronicle version (version number, commit, or branch): 

### Hardware specification

- Operating system:
- RAM:
- Cores:
- Device:

### Steps to reproduce the bug

1. 
2. 
3. 

### Errors

```
Paste error here
```

Alexandcoats:
## Linked Issues

* Closes #1228 

## Notes to Reviewer

* [x] Run chronicle and cause an error to occur (ex. by stopping the influx server). Ensure that the error is propagated to the exit code of the application.
* [x] Run chronicle as a docker image with a restart policy with retries and do above. Ensure that the docker image restarts.

### DIFF
--- a/src/bin/inx-chronicle/main.rs
+++ b/src/bin/inx-chronicle/main.rs
@@ -120,6 +120,7 @@ async fn main() -> eyre::Result<()> {
         res = process::interrupt_or_terminate() => {
             if let Err(err) = res {
                 tracing::error!("subscribing to OS interrupt signals failed with error: {err}; shutting down");
+                exit_code = Err(err);
             } else {
                 tracing::info!("received ctrl-c or terminate; shutting down");
             }
@@ -139,6 +140,7 @@ async fn main() -> eyre::Result<()> {
         res = process::interrupt_or_terminate() => {
             if let Err(err) = res {
                 tracing::error!("subscribing to OS interrupt signals failed with error: {err}; aborting");
+                exit_code = Err(err);
             } else {
                 tracing::info!("received second ctrl-c or terminate; aborting");
             }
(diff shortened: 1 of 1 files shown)


Alex6323 (approved):

Alex6323:
I already approved because the PR is easy enough to see that it will (with very high prob :trollface:) work as intended. But I'll try to  check the boxes today.

src/bin/inx-chronicle/main.rs
```diff
@@ -136,6 +140,7 @@ async fn main() -> eyre::Result<()> {
         res = process::interrupt_or_terminate() => {
             if let Err(err) = res {
                 tracing::error!("subscribing to OS interrupt signals failed with error: {err}; aborting");
+                exit_code = Err(err);
```
Alex6323, line 143: the only thing that annoys me a tiny bit here (because that should never happen :trollface: ) is that we would overwrite a previous error which would then maybe mislead the user about the root cause for the error.

Alexandcoats: Sure, but let's be honest, if you can't subscribe to interrupts you have bigger problems.

Alex6323: yeah

PR #1232, opened 2023-05-09
chore(logging): rephrase logs during shutdown

Alex6323:
I think "Shutdown successful" was confusing to people. We can circumvent this by being more accurate in logging that all our tasks have successfully resolved due to the shutdown/abort signal.

### DIFF
--- a/src/bin/inx-chronicle/main.rs
+++ b/src/bin/inx-chronicle/main.rs
@@ -127,7 +127,7 @@ async fn main() -> eyre::Result<()> {
         },
         res = tasks.join_next() => {
             if let Some(Ok(Err(err))) = res {
-                tracing::error!("A worker failed with error: {err}");
+                tracing::error!("a worker failed with error: {err}");
                 exit_code = Err(err);
             }
         },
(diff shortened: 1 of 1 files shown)


Alexandcoats:
I dunno, tasks is kind of an implementation detail. Has anyone actually complained about this?

Alex6323:
yeah, Holger in #1230 wrote:
"Chronicle should exit with a different (failure error) code so that docker restarts it when there is a hiccup with influx. It should not report as "successful shutdown" in that case."

Alexandcoats:
> "Chronicle should exit with a different (failure error) code so that docker restarts it when there is a hiccup with influx. It should not report as "successful shutdown" in that case."

This does not mean that we should change the message. He is complaining about the error, which we have fixed already.

Alex6323:
But when the error happens it still reports "Shutdown successful" atm.

Alexandcoats:
That's not necessarily wrong. We did shutdown successfully. However, I see your point. I would probably just remove "successfully" from both messages.

PR #1247, opened 2023-08-15
chore(deps)!: use `iota-sdk` and remove dependency causing security alert

### LINKED ISSUE #1246: RUSTSEC-2022-0093: Double Public Key Signing Function Oracle Attack on `ed25519-dalek`
> Double Public Key Signing Function Oracle Attack on `ed25519-dalek`

| Details             |                                                |
| ------------------- | ---------------------------------------------- |
| Package             | `ed25519-dalek`                      |
| Version             | `1.0.1`                   |
| URL                 | [https://github.com/MystenLabs/ed25519-unsafe-libs](https://github.com/MystenLabs/ed25519-unsafe-libs) |
| Date                | 2022-06-11                         |
| Patched versions    | `>=2`                  |

Versions of `ed25519-dalek` prior to v2.0 model private and public keys as
separate types which can be assembled into a `Keypair`, and also provide APIs
for serializing and deserializing 64-byte private/public keypairs.

Such APIs and serializations are inherently unsafe as the public key is one of
the inputs used in the deterministic computation of the `S` part of the signature,
but not in the `R` value. An adversary could somehow use the signing function as
an oracle that allows arbitrary public keys as input can obtain two signatures
for the same message sharing the same `R` and only differ on the `S` part.

Unfortunately, when this happens, one can easily extract the private key.

Revised public APIs in v2.0 of `ed25519-dalek` do NOT allow a decoupled
private/public keypair as signing input, except as part of specially labeled
&quot;hazmat&quot; APIs which are clearly labeled as being dangerous if misused.

See [advisory page](https://rustsec.org/advisories/RUSTSEC-2022-0093.html) for additional details.

Alexandcoats:
## Linked Issues

* Closes #1246

## Notes to Reviewer

#### API Changes
* [ ] Review the API documentation changes and confirm that it matches the actual functionality.
* [ ] Check for breaking changes in the API and matching (conventional) commit message prefix.

#### Test cases

#### Database Changes


src/bin/inx-chronicle/api/secret_key.rs
```diff
@@ -39,28 +33,24 @@ impl std::fmt::Debug for SecretKey {
...
         sk_bytes.zeroize();
-        Ok(SecretKey(secret))
+        Self(key)
     }
 
     /// Create an Ed25519 secret key from a PEM file.
-    pub fn from_file(path: &str) -> Result<SecretKey, SecretKeyError> {
-        let mut bytes = KeypairBytes::from_pkcs8_pem(&std::fs::read_to_string(std::path::Path::new(path))?)
+    pub fn from_file(path: &str) -> Result<Self, SecretKeyError> {
+        let bytes = KeypairBytes::from_pkcs8_pem(&std::fs::read_to_string(std::path::Path::new(path))?)
```
Alex6323, line 52: don't those bytes need zeroizing?

Alex6323:
Not entirely sure what lints replace it, but probably just remove the `private_in_public` lint?

Alex6323 (dismissed):

Alex6323 (approved):

PR #1249, opened 2023-08-24
Specify Hornet net config supplied by .env

trungtt1981:
Add the ability to specify Hornet net config JSON file that is supplied by the .env.
This is to avoid direct editing the file `docker\docker-compose.yml`.
So, in the .env, a new param `HORNET_CONFIG_PATH` is introduced.


docker/docker-compose.yml
```diff
@@ -84,6 +84,7 @@ services:
       # - "-c"
       # - "config_testnet.json"
       # - "config_alphanet.json"
+      - "-c ${HORNET_NET_CONFIG_JSON_FILE}"
```
Alexandcoats, line 87:
```suggestion
      - "-c ${HORNET_CONFIG_PATH}"
```

Alexandcoats (changes requested):
I think we also need to add the variable to the example `.env`

EDIT: I guess there is no example file, so maybe just add it to the README section.

trungtt1981:
> I think we also need to add the variable to the example `.env`
> 
> EDIT: I guess there is no example file, so maybe just add it to the README section.

I added a template file `.env.template` that should be OK

.env.template
```diff
@@ -0,0 +1,7 @@
+MONGODB_USERNAME=root
+MONGODB_PASSWORD=root
+INFLUXDB_USERNAME=root
+INFLUXDB_PASSWORD=password
+JWT_PASSWORD=password
+JWT_SALT=saltines
+HORNET_CONFIG_PATH="path to hornet config JSON file"
```
Alexandcoats, line 7:
```suggestion
HORNET_CONFIG_PATH=config_testnet.json
```

Alexandcoats:
Can you add it to the README as well?

trungtt1981:
> Can you add it to the README as well?

added

.env.template
```diff
@@ -0,0 +1,7 @@
+MONGODB_USERNAME=root
+MONGODB_PASSWORD=root
+INFLUXDB_USERNAME=root
+INFLUXDB_PASSWORD=password
+JWT_PASSWORD=password
+JWT_SALT=saltines
+HORNET_CONFIG_PATH=config_testnet.json
```
Alexandcoats, line 7:
```suggestion
HORNET_CONFIG_PATH=config_testnet.json

```

Alexandcoats (approved):

trungtt1981:
> Oh, you need to sign your commits, please.

what do you mean by `sign the commits` ??

Alexandcoats:
https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits

This repo requires all commits to be signed.

trungtt1981:
> https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits
> 
> This repo requires all commits to be signed.

Can I still sign the commits that are already pushed?
I realize that I use 2 different github accounts !!
If I use the github account that is same as the one creating the PR, then do I still need to sign.

Maybe, I should close this PR and then open another PR with only 1 github account for commit?

PR #1250, opened 2023-08-24
Specify Hornet net config supplied by .env

trungtt1981:
Add the ability to specify Hornet net config JSON file that is supplied by the `.env`.
This is to avoid direct editing the file docker\docker-compose.yml.
So, in the `.env`, a new param `HORNET_CONFIG_PATH` is introduced.


Alexandcoats:
This commit is still unsigned 😆

PR #1314, opened 2024-01-09
Consider expiration return addresses for ledger updates

### LINKED ISSUE #1312: Update balance endpoint to stardust
## Description

The current stardust chronicle balance endpoint returns `totalBalance` & `sigLockedBalance`, where `sigLockedBalance` does not provide much valuable information considering all the unlock conditions there are at the moment in stardust. Instead of `sigLockedBalance` I suggest adding `availableBalance` which considers all the unlocked outputs which have a time unlock condition and expired outputs with an expiration time unlock condition & the return address unlock condition matching the relevant address.

In the future, we might even want to add a `potentialBalance` that counts also for example the storage deposit unlock condition, but this needs more definition in a future balance TIP.

## Motivation

In the explorer we use the balance endpoint to display the total & available balances, and at the moment `sigLockedBalance` is considered the available balance and that is incorrect at this moment. This issue came up very visible now after the supply increase, as there are a lot of addresses that got time locked outputs that unlock on a vesting schedule, for example, the IF address https://explorer.iota.org/mainnet/addr/iota1qp8h9augeh6tk3uvlxqfapuwv93atv63eqkpru029p6sgvr49eufyz7katr?tab=Transactions, and what the explorer now displays as available balance is incorrect, because `sigLockedBalance` does not take into account unlocked outputs with a time lock condition

## Requirements

1. 
2. 
3. 

No.

### LINKED ISSUE #1313: Include outputs relative to the return address unlock condition in the history endpoint
## Description

Scenario: tx history of A
Real tx history: 
- A has 0 SMR
- A receives 40 SMR from the faucet
- A sends to B 30 SMR with an expiration of 2h, and the return address unlock condition of A
- 3h pass
- A sends 30 SMR to C

The real transaction history would then be

- ➖ `25 SMR` to C
- ➖ `30 SMR` to B (expired)
- ➕ `40 SMR`

Chronicle does not include in the tx history endpoint outputs spent by address in the return address unlock condition, so what the explorer will create bsed on the chronicle event will be

-  ➕ `5 SMR` to C ⚠️ 
- ➖ `30 SMR` to B (expired)
- ➕ `40 SMR`

The issue is that when A sends 25 SMR to C the transactions is like

- inputs
    - 30 SMR, address unlock condition B, return address unlock condition A (input 1)
- outputs:
    - 25 SMR, address unlock condition C (output 1)
    - 5 SMR, address unlock condition B (output 2)

In the transaction history chronicle is not providing the output used in the transaction as input because the address unolock condition is B, not A. 

We need to get this output in order to be able to create a correct transaction history since outputs can also be spent by the address in the return address unlock condition if conditions meet. 

## Requirements

1. 
2. 
3. 

Yes/No.

DaughterOfMars:
This PR changes how the owning address is calculated for spent outputs. Rather than assume the address unlock condition for basic/NFT outputs, the spent timestamp is checked against the expiration unlock condition, if it exists. Therefore, in the case where the output was spent on or after the timestamp, we use the return address to index the output.

UPDATE: Rolled the changes for the balance endpoint in as well for simplicity. The endpoint now returns `availableBalance` instead of `sigLockedBalance` and the calculation has been adjusted to include outputs with timelock conditions that have become unlocked, as well as those with matching expiration return addresses that are expired.

## Linked Issues

* Closes #1313
* Closes #1312 

## Notes to Reviewer

#### API Changes
* [ ] Test the ledger updates by address endpoint.

#### INX Changes

#### Database Changes
* [ ] Review migration for correctness.

### DIFF
--- a/documentation/api/api-explorer.yml
+++ b/documentation/api/api-explorer.yml
@@ -275,13 +275,13 @@ components:
         totalBalance:
           type: string
           description: >-
-            The total value held in unspent outputs owned by the given address 
-            (includes funds held in storage deposit).
+            The total value held in unspent outputs that is unlockable by the given address or currently timelocked.
+            Does not include funds held in storage deposit.
         availableBalance:
           type: string
           description: >-
-            The sum of value held in unspent outputs owned by the given address 
-            that are signature locked ("trivially unlockable").
+            The total value held in unspent outputs that is immediately unlockable at ledgerIndex by the given address.
+            Does not include funds held in storage deposit.
         ledgerIndex:
           type: integer
           description: The ledger index for which the balance calculation was performed.
--- a/documentation/api/api-explorer.yml
+++ b/documentation/api/api-explorer.yml
@@ -277,7 +277,7 @@ components:
           description: >-
             The total value held in unspent outputs owned by the given address 
             (includes funds held in storage deposit).
-        sigLockedBalance:
+        availableBalance:
           type: string
           description: >-
             The sum of value held in unspent outputs owned by the given address 
@@ -585,7 +585,7 @@ components:
     balance-example:
       value:
         totalBalance: 100000
-        sigLockedBalance: 99900
+        availableBalance: 99900
         ledgerIndex: 500000
     ledger-updates-address-example:
       value:
--- a/src/model/block/payload/transaction/output/mod.rs
+++ b/src/model/block/payload/transaction/output/mod.rs
@@ -34,7 +34,9 @@ pub use self::{
     treasury::TreasuryOutput,
 };
 use crate::model::{
-    bytify, payload::TransactionId, stringify, ProtocolParameters, TryFromWithContext, TryIntoWithContext,
+    bytify,
+    payload::{milestone::MilestoneTimestamp, TransactionId},
+    stringify, ProtocolParameters, TryFromWithContext, TryIntoWithContext,
 };
 
 /// The amount of tokens associated with an output.
@@ -142,13 +144,26 @@ pub enum Output {
 
 impl Output {
     /// Returns the [`Address`] that is in control of the output.
-    pub fn owning_address(&self) -> Option<&Address> {
+    pub fn owning_address(&self, spent_timestamp: impl Into<Option<MilestoneTimestamp>>) -> Option<&Address> {
         Some(match self {
             Self::Treasury(_) => return None,
             Self::Basic(BasicOutput {
                 address_unlock_condition,
+                expiration_unlock_condition,
                 ..
-            }) => &address_unlock_condition.address,
+            }) => {
+                if let (Some(spent_timestamp), Some(expiration_unlock_condition)) =
+                    (spent_timestamp.into(), expiration_unlock_condition)
+                {
+                    if spent_timestamp >= expiration_unlock_condition.timestamp {
+                        &expiration_unlock_condition.return_address
+                    } else {
+                        &address_unlock_condition.address
+                    }
+                } else {
+                    &address_unlock_condition.address
+                }
+            }
             Self::Alias(AliasOutput {
                 state_controller_address_unlock_condition,
                 ..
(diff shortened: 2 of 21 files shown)


documentation/api/api-explorer.yml
```diff
@@ -277,7 +277,7 @@ components:
           description: >-
             The total value held in unspent outputs owned by the given address 
             (includes funds held in storage deposit).
-        sigLockedBalance:
+        availableBalance:
```
Alex6323, line 280: also update the description?

DaughterOfMars: I didn't feel that it needed to be updated really

src/model/block/payload/transaction/output/mod.rs
```diff
@@ -142,13 +144,26 @@ pub enum Output {
...
+                    (spent_timestamp.into(), expiration_unlock_condition)
+                {
+                    if spent_timestamp >= expiration_unlock_condition.timestamp {
+                        &expiration_unlock_condition.return_address
+                    } else {
+                        &address_unlock_condition.address
+                    }
+                } else {
+                    &address_unlock_condition.address
+                }
```
Alex6323, line 165: have you considered factoring this out into a function to reduce code duplication?

DaughterOfMars: sure, why not

src/model/block/payload/transaction/output/mod.rs
```diff
@@ -142,13 +144,26 @@ pub enum Output {
 
 impl Output {
     /// Returns the [`Address`] that is in control of the output.
-    pub fn owning_address(&self) -> Option<&Address> {
+    pub fn owning_address(&self, spent_timestamp: impl Into<Option<MilestoneTimestamp>>) -> Option<&Address> {
```
Alex6323, line 147: Maybe I am misunderstanding something but isn't it the case that when the output is still unspent, but the latest milestone timestamp > expiration timestamp, then the owning address should already be the return address?

with the current logic the owner of an unspent output would still be the address in the address_unlock_condition despite we're passed the expiration.

so maybe this should just be `MilestoneTimestamp` and if the output is already spent, then it would be the `spent_timestamp` like currently, but when the output is still unspent then this should be the latest milestone timestamp.

Alex6323, line 147: Maybe I am misunderstanding something but isn't it the case that when the output is still unspent, but the latest milestone timestamp > expiration timestamp, then the owning address should already be the return address?

with the current logic the owner of an unspent output would still be the address in the address_unlock_condition despite we're passed the expiration.

so maybe this should just be MilestoneTimestamp and if the output is already spent, then it would be the spent_timestamp like currently, but when the output is still unspent then this should be the latest milestone timestamp.

documentation/api/api-explorer.yml
Alex6323: well, to me it seems a bit wrong now since `availableBalance` doesn't just consider signature locked outputs anymore, but also Timelocks and Expirations.

src/model/block/payload/transaction/output/mod.rs
DaughterOfMars: The problem is there's simply no good way for us to do this. We would have to update all unspent outputs every milestone, which is simply not performant. Spending proves definitively who the owner is, so we just update it then. Since this is only used in the explorer routes, this is fine. The only concern I have is with analytics, because we use it for active addresses, balance, and base token activity and those calculations may possibly be wrong now.

documentation/api/api-explorer.yml
DaughterOfMars: well that description is for total balance

Alex6323: well, yeah, it's just github that shows you the L277-279 because it things that would be helpful while it is just confusing here.. I was actually referring to the description of `availableBalance` in L283-284, which says:
"The sum of value held in unspent outputs owned by the given address that are signature locked ("trivially unlockable")." which is not correct anymore.

DaughterOfMars: Ahh. What do you think it should say?

DaughterOfMars:
https://chronicle-migr.stardust-mainnet.iotaledger.net/api/

This server can be used for testing the mainnet API before we commit to a migration.

Issue #1380, opened 2024-04-17
Random Chronicle crash


shufps:
We have the problem that Chronicle randomly crashes with this error:

```
inx-chronicle    | 2024-04-04T16:08:51.074357Z ERROR handle_ledger_update{slot_index=11590 created=0 consumed=0}: inx_chronicle::inx: error=missing consumed output (0x348d2dd7fba634d8b5ea52b3ae16bf1c3ac6bdd227ba786c9a004f59565aeb1e332d00000000) in slot 11590
inx-chronicle    | 2024-04-04T16:08:51.074535Z ERROR inx_chronicle: a worker failed with error: missing consumed output (0x348d2dd7fba634d8b5ea52b3ae16bf1c3ac6bdd227ba786c9a004f59565aeb1e332d00000000) in slot 11590
inx-chronicle    | Error: missing consumed output (0x348d2dd7fba634d8b5ea52b3ae16bf1c3ac6bdd227ba786c9a004f59565aeb1e332d00000000) in slot 11590
inx-chronicle    | Location:
inx-chronicle    |     /inx-chronicle/src/analytics/mod.rs:352:20
```

It might be some issue with the ordering how data is stored into the database and that there is no guarantee that INX-streams are in sync to each other.

When deleting the database, resyncing works. But it can take a couple of hours afterwards that it crashes again.

The database is then broken und recovery is not possible

PR #1381, opened 2024-04-18
Ensure transactions are finalized so outputs are available in the ledger update store

Alex6323:
## Linked Issues

It's one option that could cause #1380, but needs to be tested still since it occurs rarely only.


src/analytics/mod.rs
```diff
@@ -343,42 +346,44 @@ impl<'a, I: InputSource> Slot<'a, I> {
...
-                        output_id,
-                        slot_index: metadata.transaction_id.slot_index(),
-                    })?
-                    .clone())
-            })
-            .collect::<eyre::Result<Vec<_>>>()?;
-        analytics
-            .handle_transaction(payload, metadata, &consumed, &created, ctx)
-            .await?;
+        if let Some(TransactionState::Finalized) = metadata.transaction_state {
```
DaughterOfMars: I think I'd prefer this just because there's no bindings in the match

```suggestion
        if metadata.transaction_state == Some(TransactionState::Finalized) {
```

DaughterOfMars: Also we should move the check up to the call-site so we don't have to indent this whole fn

Alex6323: alrighty

DaughterOfMars (approved):

PR #1385, opened 2024-04-26
fix: `commitments/by-index/:index/blocks` route should return only finalized blocks

Alex6323:
Currently in the explorer route `commitments/by-index/:index/blocks` blocks are already shown, when they are just `accepted`, but not yet `finalized`. This PR generalizes the corresponding routes handler (that forwards the query to MongoDb) to additionally provide the block state. For the Explorer API specifically that means, that the handler internally sets the added `block_state` parameter to `Some(BlockState::Finalized)`.

We could also not do it like this, but instead make this mongodb query specific to finalized blocks, but then it would be specifc to the explorer api which I didn't like that much, but I am open to change it if wanted.

UPDATE: this PR was changed to only store blocks if and when they are finalized. Hence, we removed the corresponding block state analytic.


src/tangle/slot_stream.rs
```diff
@@ -43,25 +46,35 @@ impl<'a, I: InputSource> Slot<'a, I> {
...
-                None
-            };
-            Ok(BlockWithTransactionMetadata {
-                transaction,
-                block: res,
+        Ok(self
+            .source
+            .accepted_blocks(self.index())
+            .await?
+            .filter(|res| {
```
DaughterOfMars: use `try_filter` instead to simplify a bit?

PR #1388, opened 2024-05-08
fix: check commitment id

### LINKED ISSUE #1387: [Bug]: Fetching commitments by transaction id
### Bug description

Calling commitments endpoint with transaction Id instead of commitment id still resolves to a slot commitment instead of failing.
example:
```
tx id: 0x5f73f9666cc25bbbf553d71ff73034a61cb3f089214ff530c5b2909f35fce341ed680000
https://chronicle.iota2-alphanet.iotaledger.net/api/core/v3/commitments/0x5f73f9666cc25bbbf553d71ff73034a61cb3f089214ff530c5b2909f35fce341ed680000

response:
{"protocolVersion":3,"slot":26861,"previousCommitmentId":"0xfa247cd020b6a16536c4297eb06cbc6d49e9e21f3c0e3140a760b72b54de5245ec680000","rootsId":"0xd211d5cc1771de46ff023bce993c8e5a90f4101f5451d9b95d7d38abab240780","cumulativeWeight":"428823","referenceManaCost":"1"}
```

Alex6323:
Fixes #1387


Thoralf-M (approved):
Changelog entry :trollface:

src/bin/inx-chronicle/api/core/routes.rs
```diff
@@ -333,6 +333,13 @@ async fn commitment(
...
     headers: HeaderMap,
 ) -> ApiResult<IotaRawResponse<SlotCommitment>> {
+    database
+        .collection::<CommittedSlotCollection>()
+        .get_id_for_slot_index(commitment_id.slot_index())
+        .await?
+        .and_then(|id| (id == commitment_id).then_some(()))
+        .ok_or(MissingError::NoResults)?;
+
     commitment_by_index(database, Path(commitment_id.slot_index()), headers).await
```
DaughterOfMars: rather than call the database twice just to avoid a small amount of repeated code, maybe we should just fully implement this method separately from `commitment_by_index`

Alex6323: sure, if you prefer that

DaughterOfMars: yes pls 🙏

Alex6323: hmm, that does seem to require a new index though, right? :thinking:

DaughterOfMars: This is what I would do:
```rust
   let slot_commitment = database
        .collection::<CommittedSlotCollection>()
        .get_commitment(index)
        .await?
        .ok_or(MissingError::NoResults)?;

    if slot_commitment.commitment_id != commitment_id {
        return Err(MissingError::NoResults);
    }

    if matches!(headers.get(axum::http::header::ACCEPT), Some(header) if header == BYTE_CONTENT_HEADER) {
        return Ok(IotaRawResponse::Raw(slot_commitment.commitment.data()));
    }
    Ok(IotaRawResponse::Json(slot_commitment.commitment.into_inner()))
```

Alex6323: ah, sure ... that is what I had done initially. I didn't like it because of the repetition and because we have the index on the slot index, but I don't mind.

DaughterOfMars (approved):

PR #1389, opened 2024-05-13
fix: race condition for finalized blocks

Alex6323:
The problem is that we currently call `get_accepted_blocks` too early when blocks were "just" confirmed in the node software. I confirmed this by waiting for 2s before making that call, and then Chronicle was able to fetch them as finalized.

Instead of an arbitrary waiting time, this PR relaxes the block-state requirement for Chronicle to "confirmed or finalized".


Alex6323:
we found a good solution with #1395

PR #1395, opened 2024-05-13
fix(inx): wait for slots to be finalized in INX

DaughterOfMars:
# Description

Forces chronicle to wait for slots to be finalized using the node status `latest_finalized_commitment` field.

## Notes to Reviewer

#### INX Changes


src/inx/client.rs
```diff
@@ -45,17 +62,84 @@ impl Inx {
...
+                                        break;
+                                    }
+                                }
+                                None => {
+                                    return Ok(None);
+                                }
+                            }
+                        }
+                    }
+                    let commitment = inx.get_committed_slot(state.curr_slot.into()).await?;
```
Alex6323, line 132: not entirely sure I understand the call with this argument tbh ... if we assume that `state.latest_finalized_slot` got updated in the previouse loop, then it still could be `< state.curr_slot`, no? :thinking: shouldn't we call it with `state.latest_finalized_slot` instead? Or is `curr_slot` in the request always set so that can't be a problem?

DaughterOfMars: This iterates over the indexes in order, so we always get the slot with the current index. The loop above just waits until the status reports that the current index is finalized.

Alex6323: yeah, but does it really ensure that? what if `inx.listen_to_status_changes()` returns many status changes that are `< curr_slot`, or can that not happen? As far as I can see from the code at L132 `state_latest_finalized_slot` can still be smaller than `curr_slot`, and then we'ld still get a committed slot that is not finalized yet. But maybe I am overlooking smth :man_shrugging:

DaughterOfMars: it loops until it returns that the slot is finalized.

Alex6323 (approved):

